PR #19, opened 2017-11-22
Update description of the "neuroimaging" python env

wmvanvliet:
There is a new Python environment available for groups that do analysis of neuroimaging data. This one is preferred over the old "brain" environment, as all the packages are up to date.

### DIFF
--- a/aalto/python.rst
+++ b/aalto/python.rst
@@ -60,14 +60,6 @@ Some highlights include:
   - rpy2 (bridge between Python and R)
   - tidyverse
 
-If you frequently use the "neuroimaging" environment, you may consider making
-it your default Python environment by adding these lines to your ``~/.zshrc``::
-
-	# Set default python environment to "neuroimaging"
-	module load anaconda3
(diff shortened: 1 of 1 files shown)


aalto/python.rst
```diff
@@ -9,45 +9,101 @@ the generic `python page under scientific computing docs
...
+  - rpy2 (bridge between Python and R)
+  - tidyverse
+
+If you frequently use the "neuroimaging" environment, you may consider making
+it your default Python environment by adding these lines to your ``~/.zshrc``::
+
+	# Set default python environment to "neuroimaging"
+	module load anaconda3
+	export CONDA_DEFAULT_ENV="/work/modules/Ubuntu/14.04/amd64/common/anaconda3/latest/envs/neuroimaging"
+	export PATH="$CONDA_DEFAULT_ENV/bin:$PATH"
```
jabl: This is not necessarily a good idea, actually. I was just solving a problem for a user where libreoffice crashed when opening a new document. The culprit turned out to be that he had installed miniconda3 in his home directory, and added that to the default $PATH. Then libreoffice tried to use that python, and due to some incompatibility it crashed.

wmvanvliet: ah, yes. Good point.

PR #39, opened 2019-12-31
Initial draft about/ section

rkdarst:
- Eventually can replace science-it.aalto.fi
- Everything here can still change - suggestions welcome
- My main thoughts:
  - Is the basic emphasis correct?
  - This structure has the user-centric info first, with the "political
    info" second.  Is that ok?
  - Phrasing of the partnerships, etc still needs to be fixed.

I am making this a draft PR for a reason, it needs comments and
continuous development to make it workable.  There are plenty of
boilerplate sections that need expanding and revision - please
comment, but direct updates welcome too.

This goes with the name "Aalto Scientific Computing" to describe our
broader project, but that is subjec to change too.


rkdarst:
I think this is ready for review.  Stuff to review in particular (this requires focused thought):

- Does this present us the way we want?  Does it hide the official project of Science-IT too much?
- List of people involved in ASC - add more, list info for the remaining people
- Do we acknowledge all collaborators
- Science-IT board needs to be updated
- CSIT could get a dedicated subpage.
- Figures could be nicer but it's good enough for now.

rkdarst:
Enrico took a look and thought it was OK.  I have asked some more people for reviews and didn't hear back, so I will merge and we can iterate later.

PR #56, opened 2020-03-04
Two minor suggestions for "Give enough information"

bast:


rkdarst:
Great, thanks for the new link and update!

PR #63, opened 2020-03-13
aalto/jupyterhub-data: Revise page for completness and new features

rkdarst:
- Data available on Aalto linux computers by default.
- Paniikki instructions simplified.
- Motivates the big picture more.


aalto/jupyterhub-data.rst
```diff
@@ -3,34 +3,63 @@ Accessing JupyterHub data
...
+
+.. csv-table::
+   :delim: |
+   :header-rows: 1
+
+   Name                    | Path on Linux servers
+   personal notebooks      | ``/m/jhnas/jupyter/u/$nn/$username/``
+   course data             | ``/m/jhnas/jupyter/course/$course_slug/data/``
+   course instructor files | ``/m/jhnas/jupyter/course/$course_slug/files/``
+   shared data             | ``/m/jhnas/jupyter/shareddata/``
```
Sefriol: Maybe a summary of these attributes could be nice, but maybe not necessary. Other than that very good summary.

rkdarst: What kind of summary?  Do you mean what does `personal notebooks`, `course data`, etc means in the user servers?  If so, good idea... we could include the paths.

Sefriol: Yeap, something like that would probably be sufficient.

Sefriol: Maybe also a quick summary of `$course_slug`, `$nn` and `$username` would not hurt either.

rkdarst: Updated, both things added.  I added a new table for `$nn` and the like - do you think that's too much space devoted to it?

rkdarst:
@Sefriol, I think you can directly merge.  Please do so for any jupyter related stuff you would like (and really, anything else if you feel it makes sense).

aalto/jupyterhub-data.rst
Sefriol: Nope. This is fine.

PR #68, opened 2020-03-17
Add page timestamp at the bottom from git

rkdarst:
- Use extension sphinx_gitstamp to do this.  Basically everything
  works automatically, but we need to manually put it into the footer.
- This is written to not fail if the extension is not installed, to
  make this as easy to build as possible.
- Untested on readthedocs so far.


rkdarst:
@eglerean, based on your idea.  I far too long trying to figure out how after searches were unsuccessful, then it was second result on an unrelated search!

It won't work on readthedocs yet, I have to set the path to this new requirements.txt file in RTD first.

PR #72, opened 2020-03-19
reorganizes the sections

hastinarimanzadeh:
I removed the `Transferring files` section since it wasn't closely associated with how to connect to  Triton. I additionally moved the part about changing the shell closer to the exercise that alluded to it. Overall I tried to make it more compact and to the point.


rkdarst:
This is definitely an improvement, so I think we should just do it!

PR #75, opened 2020-03-24
restructures the sections

hastinarimanzadeh:
I deleted the `But what about the software?` and added to the `What's next`.
Additionally created a new section for making a module collection. 

Also expanded the `what is a module?` section, and started off the `Loading modules`section with the Python example.

Overall tried to make it more organized and to give it a sensible flow.

### DIFF
--- a/triton/tut/modules.rst
+++ b/triton/tut/modules.rst
@@ -151,7 +151,8 @@ What's going on under the hood here?
 In Linux systems, different environment variables like ``$PATH`` and
 ``$LD_LIBRARY_PATH`` help figure out how to run programs.  Modules just
 cleverly manipulate these so that you can find the software you need,
-even if there are multiple versions available.
+even if there are multiple versions available.  You can see these variables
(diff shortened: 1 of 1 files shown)


triton/tut/modules.rst
```diff
@@ -130,24 +143,28 @@ Load the latest version of Matlab as::
...
-your scripts, etc.  You could put it in your
-``~/.bash_profile``, but then it will always automatically load it -
-perhaps even if you don't expect it.  Watch out for this if you get
-un-explainable bugs - it may be best to explicitely load what you
-need.  You can load any number of modules, and there is a basic
-dependency/conflict system.
+In Linux systems, different environment variables like ``$PATH`` and
+``$LD_LIBRARY_PATH`` help figure out how to run programs.  Modules just
+cleverly manipulate these so that you can find the software you need,
+even if there are multiple versions available.
```
rkdarst:
```suggestion
even if there are multiple versions available.  You can see these variables
with the **echo** command, e.g. ``echo $PATH``.
```

PR #77, opened 2020-03-26
cosmol: various fixups

rkdarst:
triton/apps/cosmol.rst: minor fixups

   triton/apps/cosmol.rst: use -launcher slurm and -clustersimple

   - This is what people on our issue tracker use, e.g. #691.  With a
    user we tested and it seems to actually use the requested number of
    CPUs.
   - Using -c seems better than -n in this situation, we thought in our
    garage today.  It's what others do at least.

   triton/apps/cosmol.rst: fix linking the .cosmol dir

   - We need to not have trailing slash, or it expects it place it there
    (-T), not in that directory.


rkdarst:
I will assume this works, comment if not.

Issue #90, opened 2020-04-26
Server SHA256 fingerprints 


mtwest2718:
My pull request (#89) on updating the SSH page has one open issue that needs to be addressed. I wanted to add a list of the SHA256 fingerprints to the remote servers folks people will be logging into. Is this worth it or should I just delete those lines in the update?

PR #91, opened 2020-04-27
triton/: Add Triton ssh fingerprints

rkdarst:
Found with these commands:

- `ls /etc/ssh/*.pub | xargs -i ssh-keygen -l -f {}`
- `ls /etc/ssh/*.pub | xargs -i ssh-keygen -l -f {} -E md5`
- `ssh-keyscan triton.aalto.fi`

Is this a good format?

Closes: #90

### DIFF
--- a/triton/tut/connecting.rst
+++ b/triton/tut/connecting.rst
@@ -102,6 +102,8 @@ instructions to connect to Triton.
 Advanced options
 ----------------
 
+You can :doc:`verify the ssh key fingerprints <../usage/ssh-fingerprints>`.
+
 See the :doc:`advanced ssh information </scicomp/ssh>` to learn how
 to log in without a password, automatically save your username 
 and more. It really will save you time.
--- a/triton/usage/ssh-fingerprints.rst
+++ b/triton/usage/ssh-fingerprints.rst
@@ -1,0 +1,27 @@
+Triton ssh key fingerprints
+===========================
+
+ssh key fingerprints allow you to verify the server you are connecting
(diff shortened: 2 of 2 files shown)


rkdarst:
>     1. Thank you Richard for putting this together.
> 
>     2. When logging into Triton or Magi, I only ever see the ECDSA SHA256 fingerprints. Is there any reason to show the others, particularly the MD5 versions?

I'm not entirely sure, I think it depends on some priority order of client and server.  I assume that we move more towards ECDSA these days, but maybe some other platforms or older versions wouldn't use them.  I'd be interested to know more, but don't right now.  On openssh, I can make it use ssh-rsa with the option `-o HostKeyAlgorithms=ssh-rsa`.  And then, there are some thinsg offered by the server.

>     3. For users who aren't working on an Aalto IT laptop, I think it would be worthwhile to list the fingerprints for the other linux servers as well.

I'll add a link to [the aalto page here](https://www.aalto.fi/en/services/linux-shell-servers-at-aalto), good point.

>     5. What is the accepted style of the documentation? My wording feels a lot more formal than this or other pieces of the scicomp-docs.

I'm always very informal, but we haven't really thought about it.  Whatever you like.

rkdarst:
I'm sorry for taking so long to get to this... I added a link to the other page on Aalto shell servers.  That lists many different protocols, and I can't be sure that there's not some older program that still uses only older versions... I'd be interested to hear what best practices are right now...

rkdarst:
I'm merging, it's good enough for now and can be improved later.  Let me know if more is needed.

PR #93, opened 2020-05-07
Updating the serial page in Triton tutorials

hastinarimanzadeh:
I updated the text and reorganized the sections. Some extra examples of Slurm commands are added.
There are a few to-dos I'd like your input on. 

- I'd like to add some more structured rules of thumb about setting resource parameters (more specific details on how long/large jobs should be).
- Is the reference page on partitions updated (partitions, number of nodes, etc.)?
- I'd additionally like to write a simple code sample (just as in `interactive.rst`) that require some logging and I/O.

There will be further synchronization between the `Interactive jobs`, `Serial jobs`, and `Array jobs`.

### DIFF
--- a/triton/tut/serial.rst
+++ b/triton/tut/serial.rst
@@ -136,8 +136,8 @@ Examples include partitions assigned to debugging("debug" partition),
 batch processing("batch" partition), GPUs("gpu" partition), etc.
 
 Command ``sinfo`` lists the available partitions. Let's see the first 4 partitions listed
-for the sake of brevity. 
-::
+for the sake of brevity::
+
    $ sinfo | head -n 5
    PARTITION     AVAIL  TIMELIMIT  NODES  STATE NODELIST
    interactive      up 1-00:00:00      2   drng pe[1-2]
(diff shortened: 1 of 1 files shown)


triton/tut/serial.rst
```diff
@@ -2,195 +2,195 @@
...
-    $ sbatch hello.slrm
+   #!/bin/bash
+   #SBATCH --time=00:05:00    
+   #SBATCH --mem-per-cpu=100  
+   #SBATCH --output=/scratch/work/%u/hello.%j.out
+   #SBATCH --partition debug
 
-.. warning:: You must use ``sbatch``, not ``bash`` to submit the job
-   to process the ``#SBATCH`` headers and run in the background.
+   srun echo "Hello ${USER}! You are on node ${HOSTNAME}"
```
rkdarst, line 51: The `${xxx}` works, but I wonder if $xxx would reduce confusion of a new person reading this (so they don't have to wonder why curly brackets are used here, but not other places).

triton/tut/serial.rst
```diff
@@ -2,195 +2,195 @@
...
 
+A partition is a set of computing nodes dedicated to a specific purpose.
+Examples include partitions assigned to debugging("debug" partition), 
+batch processing("batch" partition), GPUs("gpu" partition), etc.
 
-Status of the jobs
-==================
+Command ``sinfo`` lists the available partitions. Let's see the first 4 partitions listed
+for the sake of brevity. 
+::
```
rkdarst, line 140:
```suggestion
for the sake of brevity::

```

PR #97, opened 2020-05-20
Updating the array jobs page in Triton tutorials

hastinarimanzadeh:
I restructured the `array.rst` documentation and incorporated the Pi script into the on-page
examples. There are still some todos and things to consider:

- The `Reading input files` section is not changed much yet. I wanted to use the `pi_aggregation.py` script for this part, so I will make the changes after the script is reviewed.
- There was an allusion to parallel array jobs that I removed. I thought it's best to have it in the parallel documentation. What do you think?
- I will add a short example for 2D sampling and `%N` use cases. 
- Perhaps an exercise regarding the use of `--ntasks` and how it differs from job array should be added?

### DIFF
--- a/triton/tut/array.rst
+++ b/triton/tut/array.rst
@@ -9,7 +9,7 @@ where there is no dependency or communication among the processes.
 Their results are thus combined as the final output, e.g. Monte Carlo simulations.
 
 In Slurm context, *job arrays* are the way to run several instances of a 
-single-threaded program, known as *embarassingly parallel* paradigm. 
+single-threaded program, known as the *embarrassingly parallel* paradigm. 
 
 Introduction
 ============
@@ -52,7 +52,7 @@ as you wish. Let's name it ``hello.sh`` and write it as follows.
    #SBATCH --array=0-15
    #SBATCH --time=00:15:00
    #SBATCH --mem=200
-   # You may put the commands after SBATCH directives
+   # You may put the commands below:
    
    # Job step
    srun echo "I am array task number" $SLURM_ARRAY_TASK_ID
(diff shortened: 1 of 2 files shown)


triton/tut/array.rst
```diff
@@ -4,249 +4,305 @@ Array jobs
...
+           ;;
+       2)  
+           SEED=22 
+           ;;
+       3) 
+           SEED=60 
+           ;; 
+       4) 
+           SEED=432 
+           ;;
```
rkdarst, line 164: This is a case where I would definitely leave it in long form, so that it can be most easily edited and you can easily compare the different lines, and keep them in sync and all.

triton/tut/array.rst
```diff
@@ -4,249 +4,305 @@ Array jobs
...
+    #!/bin/bash
+    #SBATCH -n 1
+    #SBATCH --output=/dev/null
+    #SBATCH --error=/dev/null
+    #SBATCH --array=1-4
+    #SBATCH --time=01:00:00
+    #SBATCH --mem=500
 
+    n=$SLURM_ARRAY_TASK_ID
+    iteration=`sed -n "${n} p" iterations.txt`      # Get n-th line (1-indexed) of the file
```
rkdarst, line 221: Nice, this is simpler than what I had before.

triton/tut/array.rst
```diff
@@ -4,249 +4,305 @@ Array jobs
...
+and save each output to a file. Having calculated all estimations, you could take the 
+average of all the Pi values to arrive at a more accurate estimate. An example of such
+a batch script is as follows. 
+
+.. code-block:: bash
+
+   #!/bin/bash 
+   #SBATCH --job-name=pi_estimation
+   #SBATCH --output=/dev/null
+   #SBATCH --error=/dev/null
```
rkdarst, line 142: I think it woludn't be good practice to completely ignore output/error, since if something does go wrong, there's no message and causes confusing and more debugging time for us.  Since the examples redirect the pi script output to its own json, file, we can use: `-o pi.out.log --open-mode=append`, and it puts all of the outputs to the exact same file.  Perhaps confusing, but effective at (at least) not losing any errors.

rkdarst:
I plan on taking both serial and this and directly editing and accepting them now, let me know if you happen to be working on them at the same time.

PR #101, opened 2020-06-05
Updating the GPU computing tutorial page

hastinarimanzadeh:
I tried to reorganize and update the documentation. 

I included a submission script compatible with the [CUDA helloworld](https://github.com/AaltoSciComp/hpc-examples/blob/gpu-example/gpu/helloworld.cu) in the documentation.

There are a few things I should change soon:
- Should I put all the Python scripts for the examples in the hpc-examples repo? To be cloned alongside other examples?
- Change the anaconda version for the PyTorch example.


simo-tuomisto:
Looks good, merging.

PR #109, opened 2020-06-06
scicomp/packaging-software: guide to make software easily installable

rkdarst:
- Lessons learned from HPC system administration and user support of
  researchers who need to install things.
- To review:
  - What is missing?
  - It could use more links and references
  - Is there something else that already exists that we should have
    used instead?
  - Links to recommendations about the actual coding part (which is
    what most similar pages focus on)

### DIFF
--- a/scicomp/index.rst
+++ b/scicomp/index.rst
@@ -8,6 +8,7 @@ Scientific computing
    encryption
    git
    jupyter-pitfalls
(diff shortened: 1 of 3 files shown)


rkdarst:
@simo-tuomisto, any comments on this?  It should be highly relevant to your interests

PR #111, opened 2020-06-06
triton/tut/storage: note that SMB mounting may only work from Aalto laptops

rkdarst:
- This is either a recent or upcoming change.
- Recommend requesting an Aalto computer or using VDI.


rkdarst:
@mhakala approved by IRC, next up is another improvement

PR #114, opened 2020-06-07
Revises exercises for the Interactive jobs tutorial page

hastinarimanzadeh:
I added the address to the Github repo for the exercises and added a bit more information to them.

### DIFF
--- a/triton/tut/interactive.rst
+++ b/triton/tut/interactive.rst
@@ -198,7 +198,8 @@ You can clone the repository by running
    uses up a lot of memory to do nothing.  Let's play with it.
    It's run as follows:
    ``python hpc-examples/slurm/memory-hog.py 50M``, where the
-   last argument is however much memory you want to eat.
+   last argument is however much memory you want to eat.  You can use
+   ``--help`` to see the options of the program.
 
    a) Try running the program with ``50M``.
 
--- a/triton/tut/interactive.rst
+++ b/triton/tut/interactive.rst
@@ -230,8 +230,7 @@ will be used for the rest of the tutorial exercises.
    b) Add ``srun`` in front (``srun python ...``).  Use the ``seff <jobID>``
       command to see how much time the program took to run.
       (If you'd like to use the ``time`` command, you can run
-      ``srun --mem=<m> --time=<t> bash -c
-      "time python time python hpc-examples/slurm/pi.py <iters>"``)
+      ``srun --mem=<m> --time=<t>time python time python hpc-examples/slurm/pi.py <iters>``)
 
    c) Tell srun to use five CPUs (``-c 5``).  Does it go any faster?
 
--- a/triton/tut/interactive.rst
+++ b/triton/tut/interactive.rst
@@ -206,7 +206,8 @@ will be used for the rest of the tutorial exercises.
 
    b) Run the program with ``50M`` and ``srun --mem=500M``.
 
-   c) Increase the amount of memory allocated.  How much memory can
+   c) Increase the amount of memory the Python process tries to use (not the
+      amount of memory Slurm allocates).  How much memory can
       you use before the job fails?
 
    d) Look at the job history using ``slurm history`` - can you see
--- a/triton/tut/interactive.rst
+++ b/triton/tut/interactive.rst
@@ -192,7 +192,8 @@ The scripts you need for the following exercises can be found here:
 `hpc-examples/slurm
 <https://github.com/AaltoSciComp/hpc-examples/tree/master/slurm>`__)
 You can clone the repository by running
-``git clone https://github.com/AaltoSciComp/hpc-examples.git``
+``git clone https://github.com/AaltoSciComp/hpc-examples.git``.  This repository
+will be used for the rest of the tutorial exercises.
 
 1. The program ``hpc-examples/slurm/memory-hog.py``
    uses up a lot of memory to do nothing.  Let's play with it.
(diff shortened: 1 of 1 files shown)


triton/tut/interactive.rst
```diff
@@ -190,56 +188,58 @@ It's best to use smaller values when submitting interactive jobs, and more for b
...
+   last argument is however much memory you want to eat.
+
+   a) Try running the program with ``50M``.
 
    b) Run the program with ``50M`` and ``srun --mem=500M``.
 
-   c) Increase the amount of memory allocated until the job fails.
-      What happens?
+   c) Increase the amount of memory allocated.  How much memory can
+      you use before the job fails?
```
rkdarst, line 208:
```suggestion
   c) Increase the amount of memory the Python process tries to use (not the
      amount of memory Slurm allocates).  How much memory can
      you use before the job fails?
```

triton/tut/interactive.rst
```diff
@@ -190,56 +188,58 @@ It's best to use smaller values when submitting interactive jobs, and more for b
...
 
-   d) Play around with different parameters: how much memory can you
-      use?
+   d) Look at the job history using ``slurm history`` - can you see
+      how much memory it actually used? (Try using the ``--sleep``
+      option)
```
rkdarst, line 212:
```suggestion
   d) Look at the job history using ``slurm history`` - can you see
      how much memory it actually used?
      
      Note, Slurm only measures memory every 60 seconds or so.  To make the program last longer, so that can be measured, give the ``--sleep`` option to the Python process, like this: ``python memory-hog.py 50M --sleep=60``.
```
is this the right syntax for --sleep?

triton/tut/interactive.rst
```diff
@@ -190,56 +188,58 @@ It's best to use smaller values when submitting interactive jobs, and more for b
...
-      seconds: ``time python /scratch/scip/hpc-examples/slurm/pi.py
+      seconds: ``time python hpc-examples/slurm/pi.py
       500`` and so on.
 
-   b) Add ``srun`` in front (``srun python ...``).  What changes?
+   b) Add ``srun`` in front (``srun python ...``).  Use the ``seff <jobID>``
+      command to see how much time the program took to run.
+      (If you'd like to use the ``time`` command, you can run
+      ``srun --mem=<m> --time=<t> bash -c
+      "time python time python hpc-examples/slurm/pi.py <iters>"``)
```
rkdarst, line 230:
```suggestion
      ``srun --mem=<m> --time=<t>time python time python hpc-examples/slurm/pi.py <iters>``)
```
I"m not sure about this, but raw `time` also works, it's just a different program!  I think it might be better to gloss over the differenet between `time` interactively and `time` here, than introduce something else that people might have to think about.  What do you tihnk?

triton/tut/interactive.rst
```diff
@@ -190,56 +188,58 @@ It's best to use smaller values when submitting interactive jobs, and more for b
...
+<https://github.com/AaltoSciComp/hpc-examples/tree/master/slurm>`__)
+You can clone the repository by running
+``git clone https://github.com/AaltoSciComp/hpc-examples.git``
 
-   a) Try running the program with ``50M``
+1. The program ``hpc-examples/slurm/memory-hog.py``
+   uses up a lot of memory to do nothing.  Let's play with it.
+   It's run as follows:
+   ``python hpc-examples/slurm/memory-hog.py 50M``, where the
+   last argument is however much memory you want to eat.
```
rkdarst, line 201:
```suggestion
   last argument is however much memory you want to eat.  You can use
   ``--help`` to see the options of the program.
```
I hope you can use --help - can we?

triton/tut/interactive.rst
```diff
@@ -190,56 +188,58 @@ It's best to use smaller values when submitting interactive jobs, and more for b
...
-   run like this: ``python
-   /scratch/scip/hpc-examples/slurm/memory-hog.py 50M``, where the
-   last argument is however much memory you want to eat.  (also
-   available from `hpc-examples/slurm
-   <https://github.com/AaltoSciComp/hpc-examples/tree/master/slurm>`__)
+The scripts you need for the following exercises can be found here:
+`hpc-examples/slurm
+<https://github.com/AaltoSciComp/hpc-examples/tree/master/slurm>`__)
+You can clone the repository by running
+``git clone https://github.com/AaltoSciComp/hpc-examples.git``
```
rkdarst, line 195:
```suggestion
``git clone https://github.com/AaltoSciComp/hpc-examples.git``.  This repository
will be used for the rest of the tutorial exercises.
```

triton/tut/interactive.rst
hastinarimanzadeh: Yes, we can.

triton/tut/interactive.rst
hastinarimanzadeh: I was hesitant to add this but I thought it may be good to just let it be there so that anyone can try on their own. Because we started off with `time` here, I thought it's good to leave the option to use this too.

triton/tut/interactive.rst
hastinarimanzadeh: Yes, that's it.

triton/tut/interactive.rst
rkdarst: Since I'll be teaching this tomorrow, I'll add this.  I'd rather say "look, notice the time tool is different... someday you'll figure this out" than "don't worry about this weird syntax, it doesn't matter now but it's for (reason)".

PR #120, opened 2020-06-08
aalto linux: document triton-modules

rkdarst:
- Explain ``module load triton-modules`` in several key places.


rkdarst:
No comments so merging. If anyone has comments, please correct in a follow-up.

PR #126, opened 2020-06-12
Update logo to white version, better font, and outlines

rkdarst:
- Thanks to @sayravai for these improvements
- White version
- Nimbus Sans as a font, use outlines instead of fonts.
- Adjust CSS width
- Remove old logo, it is preserved in aalto-scicomp-graphics


rkdarst:
No comments, so merging.

PR #136, opened 2020-07-07
about/cs-it: Add initial CSIT page

rkdarst:
- Draft with some basic info
- To review
  - What else do we do?
  - Is the "work for us" appropriate?
  - Add our group members to the about page, if desired.
  - Can look at the science-it page for inspiration of what else can
    go here
  - What else should be moved to scicomp.aalto.fi instead of internal
    wiki?

### DIFF
--- a/about/cs-it.rst
+++ b/about/cs-it.rst
@@ -1,0 +1,67 @@
+Computer Science IT
+===================
+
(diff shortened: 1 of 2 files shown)


rkdarst:
We talked and I heard this looked good, so merging.  @jkotimak

PR #137, opened 2020-07-08
about/join: Add description of "staff network"

rkdarst:
- Separate from HPC and RSEs, we can have some sort of staff network.
  Document this (even though we don't really know what it will be).

- To review: does it sound good?  Is it a good start?  It will have to
  be iterated once people are back, and make sure it's what we want to
  do.

### DIFF
--- a/about/join.rst
+++ b/about/join.rst
@@ -25,6 +25,40 @@ departments which provide joint staff with us or support the :doc:`RSE
 program </rse/index>`, in addition to a greater share of resources.
 
 
+Staff network
+-------------
+
+*There is no Aalto Scientific Computing, just people who want to make
+computing better.*
+
+There is no "Aalto Scientific Computing" on paper, only different
+teams that work together to help researchers better than they could
+alone.  We invite interested staff to join our community, :doc:`help
+sessions </news/garage>`, infrastructure development, etc.  This
+program is just being developed (as of 2020), but it roughly includes:
+
+* Participation in admin meetings to help us develop infrastructure
+  (e.g.  Triton) in the best way for your users
+
+* Co-maintenance of infrastructure (for example, your unit's special
+  software) on Triton and in out automated software deployment
+  systems.
+
+* Networking, to learn how to solve your users' problems more
+  efficiently.
+
+* Continual professional development
+
+* This is not just for IT support or administrative support, but
+  high-quality research support that connects all aspects of work.
+
+This does not replace local support, it just makes it more powerful.
+
+.. todo::
+
+   How to take part.
+
+
 Triton: computing and data storage resources
 --------------------------------------------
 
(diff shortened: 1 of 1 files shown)


rkdarst:
I agree with both "it's hard to say no" and "it's hard to get people involved" - I've been wondering just *how* to get people involved and make them feel useful.  We have Triton, so it's easy to rally around that.  We have the automated builder system, which might be useful to others (though Linux only...).

Maybe it could also get a name that indicates its purpose better?  SciComp Staff Network?  SciComp Staff Group? ... ?

Should this be the same as the poweruser group?

By the way, I didn't see your suggestions.  Did you try to leave them using the suggest feature, because I don't see anything there, nor any new commits...?

about/join.rst
```diff
@@ -25,6 +25,40 @@ departments which provide joint staff with us or support the :doc:`RSE
 program </rse/index>`, in addition to a greater share of resources.
 
 
+Staff network
+-------------
+
+*There is no Aalto Scientific Computing, just people who want to make
+computing better.*
+
+There is no "Aalto Scientific Computing" on paper, only different
```
eglerean, line 34:
```suggestion
You might be a department IT staff member, a lab engineer, a skilled postdoc or a doctoral candidates who helps other researchers with their technical/computational challenges. Why not joining forces and join our network of helpers? There is no "Aalto Scientific Computing" on paper, only different
```

rkdarst: Just what I was looking for, thanks!

rkdarst:
Idea from a chat: staff network can be about lobbying for overall continual training and the concept that yes, you need to improve skills.  So two points to staff network:
- rally around infrastructure
- rally around training
can we make a story about this?

rkdarst:
Also "if you struggle to help and reach people, you can reach more by using our platform" "we can also provide more eyes and help with doing so."

about/join.rst
```diff
@@ -25,6 +25,40 @@ departments which provide joint staff with us or support the :doc:`RSE
 program </rse/index>`, in addition to a greater share of resources.
 
 
+Staff network
+-------------
+
+*There is no Aalto Scientific Computing, just people who want to make
+computing better.*
+
+You might be a department IT staff member, a lab engineer, a skilled postdoc or a doctoral candidates who helps other researchers with their technical/computational challenges. Why not joining forces and join our network of helpers? There is no "Aalto Scientific Computing" on paper, only different
```
rkdarst, line 34:
```suggestion
You might be a department IT staff member, a lab engineer, a skilled postdoc or a doctoral candidates who helps other researchers with their technical/computational challenges. Why not joining forces and join our network of specialists? There is no "Aalto Scientific Computing" on paper, only different
```
helpers -> specialists (makes us all seem more equal)

PR #142, opened 2020-07-21
aalto/jupyterhub: Direct link to courses

rkdarst:
- Made possible by jupyterhub/jupyterhub#3013

### DIFF
--- a/aalto/jupyterhub.rst
+++ b/aalto/jupyterhub.rst
@@ -264,9 +264,10 @@ FAQ and bugs
 
 * **I don't see the assignments for my course**.  There are different
   profiles you can start, and you can't tell which profile you have
-  started.  Go back to the hub control panel and stop your server
-  (look for the menu item to do this), and then start the profile for
-  your course.
+  started.  Go back to the hub control panel and restart your server.
+  To be more precise, click the "Control Panel" in the upper-right
+  corner, then click "Stop my Server", wait a little bit, then click
+  "Start My Server" and choose the profile for your course.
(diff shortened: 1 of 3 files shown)


aalto/jupyterhub.rst
```diff
@@ -262,6 +262,12 @@ FAQ and bugs
   and eventually needs to be cleaned up - after the respective course
   is done.
 
+* **I don't see the assignments for my course**.  There are different
```
Sefriol, line 265: Maybe this should be little more precise on what to do. Either some direct instructions on how to change spawners (i.e "On the upper right corner, click "Control Panel" next to "Logout", you should arrive to new page with either one or two buttons. First click "Stop my Server", wait a little bit and click "Start My Server" and choose a different spawner")

Or just step by step guide with pictures:
image
image
etc.

rkdarst:
good point - I added a more detailed description of changing the profile.  I didn't add screenshots yet, because classic notebooks and jupyterlab are different.  Sometime later, we should consider adding a bunch of screenshots (maybe once it is fully jupyterlab?).

Sefriol:
> good point - I added a more detailed description of changing the profile. I didn't add screenshots yet, because classic notebooks and jupyterlab are different. Sometime later, we should consider adding a bunch of screenshots (maybe once it is fully jupyterlab?).

Yeah, was thinking the same. JupyterLab integration is probably the point when we should look into this.

rkdarst:
thanks for the reviews!

PR #143, opened 2020-07-28
scicomp/zen-of-scicomp: Add a section introducing the second half

rkdarst:
- Explains why we have a list of different topics
- To review: clarity and text (whole thing could take review, too, but
  that's sfor another day)


rkdarst:
@bast, want to review?  Seems I can't assign you in github.

bast:
I like it!

PR #145, opened 2020-07-29
triton/usage/smallfiles: add an example of using Zipfiles from Python

rkdarst:


rkdarst:
@simo-tuomisto, this is relevant to your interests.  This isn't a perfect strategy for dealing with many small files, but it's something that works in simple cases.  What do you think?

simo-tuomisto:
Nice idea, but unfortunately it wont work. Here is an example in a [gist I made](https://gist.github.com/simo-tuomisto/2a75f4a63997bdd669d19b60f00bd4b9). In the situation the data is some random numpy pickles. Looking at the output it is clear that using zips in this fashion will:
- make the code slower
- make the code create more IO

The reason for this is most likely that [zlibs' DEFLATE-algorithm](https://www.euccas.me/zlib/) uses a smaller buffer than efficient file system loads (such as loading a pickle) and thus to load a file one needs to do more reads to the filesystem. The example case if of course a quite ideal for data loading, as the data is in binary blobs, but the same will apply for other cases as well.

This is also the reason why one should copy the zip to local file system on compute nodes before doing the unzip: filesystem copy commands will read the zip-file in bigger chunks than unzip. This moves the filesystem load of unzipping from the shared drive to the local drive.

rkdarst:
I thought it would be a bit worse but OK enough for archival purposes where the data doesn't need to be accessed that often, but this is just ridiculous.  (basically, not during calculations but to get rid of many files after that stage)  Oh well...

Do you have any recommendations for this kind of access pattern: basically minimal mapping of bunch of small files to one container format suitable for random access?  I'd think there should be something usable like this, without having to copy to local storage (which gets a bit much for postprocessing on e.g. frontend node).

PR #146, opened 2020-07-29
triton/usage/smallfiles: Add link to h5py quickstart

rkdarst:


simo-tuomisto:
We should definitely have a page about h5py, but I'm not certain if small files instructions-page is the correct place to have a link to h5py. HDF5 is notoriously bad when it comes to doing reads from all within the dataset and using multiple datasets [is usually very slow](https://github.com/h5py/h5py/issues/994).

HDF5 is at its best when dealing with large data that can be read in a contiguous fashion. [Chunking the data](https://docs.h5py.org/en/stable/high/dataset.html#chunked-storage) will improve random reads a bit, but as each chunk will need to be loaded into memory, random read is still not the recommended way of access.

PR #164, opened 2020-08-29
_static/theme_overrides: Improve web accessibilty by adjusting contrast

rkdarst:
58fb44b

- The reporting tool said that there was not enough contrast.  The
  upstream sphinx_rtd_theme issue is here:
  https://github.com/readthedocs/sphinx_rtd_theme/issues/971

- I decided to explore some myself and see if I could improve it some
  even in advance.  This is the results of that, and I expect this to
  be removed once the upstream is improved.

- I tried to make minimum changes, basically increasing the color
  saturation of several elements.  This isn't everything, just the
  most common things that are on almost every page.  More work is
  needed, especially on colored code blocks.

- The visible effects are rather minimal, but I personally think it
  makes it a bit easier to read, especially late at night...

- Note: I am not a web designer, and I would not recommend anyone to
  use this directly without trying to improve it themselves.  In
  particular, is the theme still consistent overall?

Specific actions:
 - Increane link text saturation
 - Increase literal text saturation
 - Increase footer contrast

- Tested with https://wave.webaim.org/

f1e6bb1

   - Further adjust link vs body contrast
   - Also work on contrast vs some of the shaded backgrounds

dbb029b

   - Adjust pygments themes

8d94ab5

   - Make the admonition title backgrounds darker.
   - These are very long CSS selectors, copied from what the current
     theme uses.

a822bb5

   - Make the sidebar background darker


rkdarst:
This should not be merged, I switched to a module-based solution that can be used in other sites, too.

rkdarst:
Replaced by #166.

PR #167, opened 2020-09-02
Initial draft of RSE procedures pages

rkdarst:
- This is a first attempt at some user-facing instructions on how to
  use our services.
- One, for how to request our services (link to Gitlab issue tracker)
- Two, how we work (what to expect, etc), based on the UCL guide I
  found.  The point is to be a longer thing to let people know what to
  expect when using our services.
- To review: everything, I haven't put much thought into this.  I
  think we shouldn't try to make it perfect right now, but presentable
  and then continue to iterate later.

PR #168, opened 2020-09-02
rse/procedures/time-tracking: Suggestion for RSE time mgmt via Gitlab

rkdarst:
- Problem: we need to track time somehow, so that we can tell each
  deparment how much benefit they are getting, and how much funding
  each group is providing and which department's funding it replaces.
- This is an initial light suggestion for how we can manage RSE
  projects using Gitlab.  Gitlab isn't exactly the right platform, but
  it is open and the same we should be using for other stuff.
- We would then closely manage the projects using labels and other
  scripts using the Gitlab APIs, to produce reports.
- To review: is this even a good idea?

### DIFF
--- a/rse/procedures/index.rst
+++ b/rse/procedures/index.rst
@@ -35,3 +35,5 @@ For ourselves
 -------------
 
 .. toctree::
+
+   time-tracking
--- a/rse/procedures/time-tracking.rst
+++ b/rse/procedures/time-tracking.rst
@@ -1,0 +1,32 @@
+Time tracking
+=============
+
+Unfortunately (fortunately?), we have to track our time some, in order
+to justify the benefits of what we do.
+
+Time tracking is done through Gitlab, within the issue opened for each
+"project".
+
+Gitlab commands:
+
+* Use these within the issue as a comment, to control the time
+  allocation.
+* ``/estimate NNw`` - estimate total time a project make take.  Used as
+  soon as possible at beginning of a project, can always be updated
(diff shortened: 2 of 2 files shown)


rantahar:
Time tracking is definitely a good idea, it should help a lot when justifying funding to departments or possibly making the case for expanding the group. It is also useful for learning project management: at first the guess will not be accurate but it will get better.

Time spent on a projects could be seen directly as a measure of the service the departments or groups receive, but it might be good to track other things as well.

In another group we've been tracking
 * papers with acknowledgement / authorship
 * enabled research projects (would not have been possible without RSE consultation)
 * contributions to open source software 
 * improvements in speed, efficiency, MPI scaling (saved node hours)

rkdarst:
Thanks!  These are good points.  I'm adding these ideas to the other
PR #167, where they need to be mentioned first.  But I'll add them
here too.

So, given all of these things you track, how did you do it at your old
group?  A simple way would be a massive spreadsheet, but that quickly
gets out of hand.  In our case, we expect more short or medium term
projects, which makes the overhead of tracking these things (and time
billed to different units) really a lot.

rantahar:
The group has a private git repository with a separate folder for each project or class of projects. They contain individual reports from each project, including the outcomes and time tracking. These do need to be somewhere, but there should be a summary somewhere as well. The last I've seen there were 20 folders in that repository.

I expect a simple spreadsheet would be sufficient for a couple of years, provided a more complete report exists in a separate repository. Of course it could be a database instead.

PR #171, opened 2020-09-02
scicomp/practical-git-prs: PR descriptions and CI checks

rkdarst:
- Talk about PR decriptions and what should go in them
- Mention CI checks
- To review: not much I think, general sanity check.

### DIFF
--- a/scicomp/practical-git-prs.rst
+++ b/scicomp/practical-git-prs.rst
@@ -119,10 +119,12 @@ Pull request description
 ~~~~~~~~~~~~~~~~~~~~~~~~
 - These days, I (rkdarst) tend to write my initial PR message into my
   commit, then ``git-pr`` will use that when I push.  This also stores
-  it permanent.
+  the description permanently in the git history.
 - There is also the concept of "pull request templates" within
-  Github/Gitlab.  (After fast PRs via ``git-pr`` I really don't like
-  this being required)
+  Github/Gitlab.  (They can keep changes organized, provide
+  checklists, and keep things moving.  But after fast small PRs via
+  ``git-pr`` I really don't like this being required for small
+  changes where I can write the important aspects myself.)
 - What should go in a description:
 
   - Why are changes being made?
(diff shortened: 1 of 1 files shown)


scicomp/practical-git-prs.rst
```diff
@@ -115,6 +115,28 @@ How do you make a pull request
 .. _gh-pr: https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/proposing-changes-to-your-work-with-pull-requests
 .. _gl-mr: https://docs.gitlab.com/ee/README.html#merge-requests
 
+Pull request description
+~~~~~~~~~~~~~~~~~~~~~~~~
+- These days, I (rkdarst) tend to write my initial PR message into my
+  commit, then ``git-pr`` will use that when I push.  This also stores
+  it permanent.
```
Gehock, line 122: I'm not sure what's "it" that's being saved permanently here

scicomp/practical-git-prs.rst
```diff
@@ -115,6 +115,28 @@ How do you make a pull request
...
+- There is also the concept of "pull request templates" within
+  Github/Gitlab.  (After fast PRs via ``git-pr`` I really don't like
+  this being required)
```
Gehock, line 125:
```suggestion
  Github/Gitlab.  (After fast PRs via ``git-pr`` I really don't feel like
  these are required)
```

Gehock (changes requested):

scicomp/practical-git-prs.rst
Gehock, line 125: I would argue that pull request templates can be beneficial for projects which have a large number of contributors, either within the team or externally. A good example of that would be the A+ LMS: https://github.com/apluslms/.github/blob/master/pull_request_template.md

Then again, that might be slightly out of scope for this document.

rkdarst: This is a good point, I was too opinionated there.  The templates do have a good purpose, especially for large feature changes.  My hesitation is biased because I have been using git-pr for making very many small changes, which (for example) that A+ template is too much.  (say, fix a typo, fix docs).  I think that these can be done quickly writing a free-form description (including the important things from the template).

scicomp/practical-git-prs.rst
rkdarst: Fixed, it referred to the pull request description: since it's the same as the commit message, it gets recorded in the log.

rkdarst (commented):
Thanks - these were good points, I pushed a new commit to update it.

scicomp/practical-git-prs.rst
Gehock: I think after the changes this looks very good, now the benefits of both tactics get mentioned

Gehock (approved):

PR #181, opened 2020-09-14
rse/procedures/job-descriptions: Initial career ladder

rkdarst:
- Write down an initial career ladder idea.  This is not so much
  because it is needed for what it is (it is pretty standard), but to
  give us something to discuss when planning the program.

- To review: everything, especially the content - though it is just a
  draft.


eglerean:
Looking good. I let somebody else to read and press the merge button.

PR #182, opened 2020-09-15
editing the news page to make it more uptodate reflect daily news

eglerean:
I would like to update the /news/index.rst daily to reflect news/upcoming events or other things worth mentioning.

I am unsure about some of the syntax (will the toctree work?) but hopefully it's quick to fix by who reviews this :)

### DIFF
--- a/news/index.rst
+++ b/news/index.rst
@@ -2,12 +2,12 @@
 News
 ====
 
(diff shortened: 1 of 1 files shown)


rkdarst:
This is a good idea, we can start here and advance it later.
I will probably not be able to remember to put something else each day, but can help to think of things.

PR #183, opened 2020-09-15
updating the index so that also the news text is shown

eglerean:


rkdarst:
This is really interesting... I wish the news items could be somehow offset, but without that it's still interesting.

I noticed that, when you have the same pages twice in the toctree, they will both be open.  that's kind of weird - I'll see what I can do after the course.

index.rst
```diff
@@ -157,3 +152,7 @@ problem is missing information, and mistakes can always be fixed.
 
 * :ref:`genindex`
 * :ref:`search`
+
+.. raw:: html
+
+   <a class="twitter-timeline" data-width="700" data-height="480" data-theme="dark" href="https://twitter.com/SciCompAalto?ref_src=twsrc%5Etfw">Tweets by SciCompAalto</a> <script async src="https://platform.twitter.com/widgets.js" charset="utf-8"></script>
```
rkdarst, line 158: I normally try to prevent going to scicomp.aalto.fi from loading any external, tracking resources (which is why I haven't embedded any youtube videos yet).  Of course it would be really nice to have this, it would basically be our news feed.  Any ideas how to have it without hotloading from twitter?

eglerean: That is the "official" way (according to Twitter of course). I can search if there are sphinx extensions to embed tweets.

eglerean:
> This is really interesting... I wish the news items could be somehow offset, but without that it's still interesting.
> 

I am not sure what you mean. In my local build I see the bullet points of news item aligned with other bullet points (e.g. above "welcome students")


> I noticed that, when you have the same pages twice in the toctree, they will both be open. that's kind of weird - I'll see what I can do after the course.

index.rst
eglerean: I tried https://pypi.org/project/sphinxcontrib.twitter/ but the result is the same (embeds the widget from Twitter)

eglerean:
ok this is ready to review. I deleted the toctree inside news to avoid messing the main toctree. I have implemented the news archive using include + start/end line. If that works well I will update the news/index.rst later to add more retroactive archive content.

index.rst
```diff
@@ -26,11 +26,14 @@ other friends.  :doc:`You can join us </about/join>`.
...
-   :maxdepth: 2
+.. include:: /news/index.rst
+   :start-line: 3
+   :end-line: 7
 
-   news/index
+.. toctree::
+   :maxdepth: 1
 
+   news/index.rst
```
rkdarst, line 36: Not important, but `.rst` isn't generally needed when linking files internally (it always finds them)

news/index.rst
```diff
@@ -2,12 +2,13 @@
...
+
+*15/09/2020* – Our course on `"Python for Scientific Computing" <training/scip/python-for-scicomp>`__ is ongoing, next session is tomorrow. It can also be watched `live on CodeRefinery Twitch <https://www.twitch.tv/coderefinery>`__ if you did not have time to register. 
 
-.. toctree::
-   :maxdepth: 1
 
-   ../training/scip/python-for-scicomp
-   ../training/index
-   ../help/garage
-   ../training/scicomp-announcements-maillist
```
rkdarst, line 13: I think maybe we should go back to the old toctree here... I'll find it and push to the branch.

rkdarst (approved):
Very good solution... I'll push the toctree update and then go.  Of course that old toctree is perhaps obsolete?

PR #185, opened 2020-09-19
news/garage: Redirect /news/garage/ to /help/garage/

rkdarst:
- This removes the weird toctree effect where navigating to
  /help/garage makes you end up in /help/garage.

- It uses raw HTML to redirect it, which I think isn't ideal but
  simple and seems to work locally, and I hope redirects search
  engines to the right place.

- There are HTTP redirects in readthedocs, but then we stay locked in
  to readthedocs.

- Review: consider if this is a good idea and if there is a better
  Sphinx option.

  - Update: I just saw
    https://github.com/wpilibsuite/sphinxext-rediraffe, which is
    released less than a month ago, which is why I didn't see it
    before.  Perhaps consider that instead of this.

### DIFF
--- a/news/garage.rst
+++ b/news/garage.rst
@@ -2,7 +2,10 @@
 Scicomp garage
 ==============
 
-.. seealso:: This page has been moved to :doc:`/help/garage`.  Below
-	     is an included for your convenience.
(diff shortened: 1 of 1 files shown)


eglerean:
I think this is fine with the raw html. Most likely it is the only case where we need such redirects.

PR #191, opened 2020-09-26
Add support for a "news area" on the sidebar

rkdarst:
- We had been exploring ways to have news in the sidebar.  I had a
  dream that pointed me the right direction, and this is the basic
  implementation.
- This is a draft but shows the basics.
- The actual items need to be generated dynamically from somewhere or
  the other, perhaps from a tweets/ directory connecting to the
  tweeting github action?
- Need some CSS theming.

### DIFF
--- a/_templates/layout.html
+++ b/_templates/layout.html
@@ -1,0 +1,13 @@
+{% extends "!layout.html" %}
+
+{% block menu %}
+{{ super() }}
+
+
+{% for news in news_items %}
+<div class="news-item">
+  {{ news }}
+</div>
+{% endfor %}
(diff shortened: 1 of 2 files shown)


eglerean:
I comment on this: I have tested this and I think it makes the left hand navigation bar cluttery. Maybe it is fine if we stick with manual news on top of homepage so I am not sure if we need this or if somebody wants to give it a second look.

eglerean:
I am not sure if we need this afterall. The sidebar is quite crowded already and having news there makes it even more busy. Should we just close this?

PR #193, opened 2020-09-26
Misc typo fixes and a link to info about guru

Gehock:
I'm not sure if the syntax for adding a link is correct, would like somebody to double-check that

### DIFF
--- a/aalto/jupyterhub.rst
+++ b/aalto/jupyterhub.rst
@@ -141,7 +141,7 @@ separate from your data).  Your actual data is yours only and you are
 responsible for it.  We do not access your data, but when necessary
 for the operation of the system, but we use and may look at file
 metadata such as permissions, timestamp, filename (``stat filename``).
-Your ``/notebooks`` directory may be deleted once your have been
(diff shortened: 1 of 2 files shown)


aalto/jupyterhub.rst
```diff
@@ -285,7 +285,8 @@ to give Github all our information, so you can also send tickets to
 CS-IT.
 
 Students and others who have difficulty in usage outside of a course
-can contact CS-IT via the guru alias.
+can contact CS-IT via the :doc:`guru <jupyterhub-instructors#more-info>`
```
Gehock, line 288: Didn't have a test environment available, haven't tested if this is the correct syntax

rkdarst:
Sorry for not doing this earlier... in Sphinx there are doc links directly to sections (how can you know ensure it will always work, if there is no semantic link?).  So, you label it with a unique reference, and then can refer to that.  As an advantage it'll stay working if the section title changes.  Though, maybe headings become automatic refs?  Not sure though.

(I think section names also become anchors and linkable from the user perspective, but you can also ref them and then they also get a permenant anchor.)

PR #212, opened 2020-10-25
adding accessibility statement

eglerean:
I have added an accessibility statement. The statement is based on a similar statement by fairdata.fi https://www.fairdata.fi/en/fairdata-fi-accessibility/

I ask @rkdarst to review since he started the work on improving accessibility earlier in the summer. I also ask @mhakala to review since I have added his work phone contact to the webpage.

### DIFF
--- a/_templates/footer.html
+++ b/_templates/footer.html
@@ -9,6 +9,6 @@
   This site is made for Aalto Scientific Computing,
   but we <a href="https://github.com/AaltoSciComp/scicomp-docs">welcome
   contributions</a> that make it more useful to others, too.  Built
-  with <a href="https://www.sphinx-doc.org/">Sphinx</a>. This website is partially conformant with the <a href="https://www.w3.org/WAI/standards-guidelines/wcag/">Web Content Accessibility Guidelines (WCAG)</a> level AA. <a href="/accessibility/">Read more about the accessibility of scicomp.aalto.fi</a>.
+  with <a href="https://www.sphinx-doc.org/">Sphinx</a>.
 </p>
 {% endblock %}
--- a/_templates/footer.html
+++ b/_templates/footer.html
@@ -9,6 +9,6 @@
   This site is made for Aalto Scientific Computing,
   but we <a href="https://github.com/AaltoSciComp/scicomp-docs">welcome
   contributions</a> that make it more useful to others, too.  Built
-  with <a href="https://www.sphinx-doc.org/">Sphinx</a>.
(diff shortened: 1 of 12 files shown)


mhakala:
I'd prefer having a common (Aalto central) phone number there. Having personal work phones in these kind of document is generally bad idea. If ok, please change this to 09 47001.

eglerean:
> I'd prefer having a common (Aalto central) phone number there. Having personal work phones in these kind of document is generally bad idea. If ok, please change this to 09 47001.

Good point! Phone number updated.

eglerean:
@rkdarst if you'd rather have the "Accessibility" link on the footer, just let me know and I can move it there.

rkdarst:
I think the bottom would be better.  If you can think of other ways to reduce the number of things on the left sidebar, I am always interested!

eglerean:
hi @rkdarst, I have moved the accessibility link from left hand menu to footer and tested it (why the footer wants links to be in html btw?).

If you think it's fine, please merge into master.

eglerean:
Can I merge this?

rkdarst:
I was still thinking about reducing some text on the main page: there
is a lot of stuff that doesn't seem to provide value, beyond a single
link that says "Accessibility".  If it was something more useful, like
"How we want to make our community accessible" (instead of website
which is self-evident), then I would do more.  So, I would remove text
from main page and either have "Accessibility" in footer, or if it's
allowed then just a single link in the toctree under "About".

(in fact, one of my ideas was to troll people by "required
accessibility statement" which says nothing useful, and also the
"actual accessibility goals" where we talk about how we hope that
everyone can actually use our services - as aalto.fi teaches us, web
accessibility != usefulness.)

eglerean:
I agree with you that this is not useful for our users, but this is just to comply with a law. We need all that text as it is as in https://www.fairdata.fi/en/fairdata-fi-accessibility/ since that’s what Antti recommended. If you think some of that info is redundant, I can meet with Antti again and confirm that he’s fine with whatever final version of the statement we have as soon as it’s published.

_templates/footer.html
```diff
@@ -9,6 +9,6 @@
   This site is made for Aalto Scientific Computing,
   but we <a href="https://github.com/AaltoSciComp/scicomp-docs">welcome
   contributions</a> that make it more useful to others, too.  Built
-  with <a href="https://www.sphinx-doc.org/">Sphinx</a>.
+  with <a href="https://www.sphinx-doc.org/">Sphinx</a>. This website is partially conformant with the <a href="https://www.w3.org/WAI/standards-guidelines/wcag/">Web Content Accessibility Guidelines (WCAG)</a> level AA. <a href="/accessibility/">Read more about the accessibility of scicomp.aalto.fi</a>.
```
eglerean, line 12: We can just leave the bit that says "Read about the accessibility of scicomp.aalto.fi" if it takes too much space in the footer.

eglerean (commented):
We can just leave the bit that says "Read about the accessibility of scicomp.aalto.fi" if it's too long in the footer.

PR #215, opened 2020-11-05
rse: Update documents with information about overhead handling

rkdarst:
- When changing to a project, we have to charge only that work to that
  project.  So we need a different way of handling the program
  administration costs.
- Update various documents to reflect this
- Review: general check / does anything obviously not make sense?  It
  will probably have to be iterated anyway, the process isn't fully
  known yet.


rkdarst:
I double-checked this and since I'm likely the expert here, I will merge it.

PR #217, opened 2020-11-06
training/scip/data-analysis: Added link to the special assignment.

simo-tuomisto:
Added new links that point to the special assignment in course repository.

### DIFF
--- a/training/scip/data-analysis.rst
+++ b/training/scip/data-analysis.rst
@@ -99,6 +99,9 @@ hours correspond roughly to 1 ECTS. Students who wish to get a
 certificate should hand in the special assignment and participate to
 at least 3 of 4 lectures.
 
+Instructions on how to obtain the special assignment can be found at the
+Materials-section.
+
 **Preparation**
 
 Software installation is done via conda. Installation instructions are provided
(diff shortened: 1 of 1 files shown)


rkdarst:
This looks good.  I was looking at the exercise notebooks and here is what I noticed:

* Make it clear what we expect for exercise 2.  For example, in 1.1 and 2.1: what if they do the task, but don't print out the results?  Do we have to go digging for it in the variables? I'm not sure the best solution, but it could be a combination of:
  * Say, in the overall instructions, to make sure to clearly print out each answer with a markdown heading, and that all answers must be clearly printed out (we will not modify the notebook)
  * Create template cells that show the heading, e.g. `## Answer 2.1: preprocessed table` and then expect them to print the transformed table right below.  Same for each other question.  Basically, optimize for making the grading as fast as possible.
* The bash_prompt_counts download doesn't seem to work.

rkdarst (approved):

simo-tuomisto:
I was planning on running the notebooks and seeing the variables after the code has been run.

The reason why there's no model answers is that the special assignment should be a bit more free form compared to the course exercises and students should make calls on how to do the operations. Setting locked solutions will discourage that.

This might slow down the grading a bit, but I want the special assignment to be something that can be graded not only by giving points, but also by giving feedback on whether the solution was good from the standpoint of the course material (modular, optimized etc.). This won't affect the grading, but IMO is a part of the course's general idea.

Link to dataset fixed. Thanks for noticing!

rkdarst:
Hm, perhaps I could rephrase my idea to be "make sure that each the answer to each question can be clearly found in the notebook, preferably labeled with a markdown heading.  Each solution output should be clearly displayed somehow (for example, the intermediate tables we ask you to create - just the head is enough)."

PR #218, opened 2020-11-09
rse: add case studies

rkdarst:
- Someone suggested that "case studies" would be a good way to
  introduce what the program can do, rather than long texts.  Add them
  for researchers and units.
- Review: check if these are representative of what we do and somewhat
  correct in wording.

### DIFF
--- a/rse/researchers.rst
+++ b/rse/researchers.rst
@@ -11,6 +11,31 @@ To request a service, contact Aalto Scientific Computing or `answer
 our survey and request to be contacted
 <https://forms.gle/wxnCpCGxdUfGGqfw6>`__ this form.
 
+.. admonition:: Case study: preparation for publication
+
+   A group is about to publish a paper about a method, but their code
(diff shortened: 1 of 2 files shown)


wmvanvliet (commented):
Love the first two use cases. For the third, I think RSEs can do more than just introduce Git. We can also help structure the codebase. Refractor stuff that is used across projects into into a lib with a good API. Etc.

eglerean:
Added a case about sustainability of old projects

rantahar:
I like all of these and they cover most important cases. The one @wmvanvliet mentions, building a library/api out of code from one or more projects, could be added as a separate use case.

wmvanvliet (approved):
I like the use cases this way!

rkdarst:
I will merge this so I can send it to someone, we can always continue to improve.

I was also wondering about the balance between case studies and the list below (to avoid having the page become too long), but it seems good now.

PR #220, opened 2020-11-11
Reword parts of the request and implementation pages

rantahar:
Reworded a couple of sentences. I found these a bit confusing.

Review:
 - Is what I wrote what you intended?
 - Is it better?


rantahar:
Adding a fix to the implementation page. The original sentence does not work, maybe this is what you meant

rkdarst (approved):

PR #221, opened 2020-11-11
Attempt to clarify project funding replacing departments share

rantahar:
Review
 - Is this what you meant
 - language


rkdarst:
Yes, this looks like a better description.  Thansk!

The Sphinx warning really confused me, since it said line 4 but that's just the start of the paragraph.  Somehow I hadn't seen that before...

PR #232, opened 2020-11-25
Add Zulip page

piiroim1:


aalto/zulip.rst
```diff
@@ -0,0 +1,75 @@
...
+
+* AzureAD 
+
+  * This is Aalto Login and should be enabled 
+
+* Email 
+
+  * This allows users to register using an email address 
+  * If you enable this, make the chat ``invitation only`` as described in 'Invitation settings'
+  * We cannot allow random people or bots to register freely
```
rkdarst, line 68:
```suggestion
  * We cannot allow random people or bots to register freely  
  * Unfortunately, there is no option for "Anyone with an Aalto account can register without invitation, but you can invite others with an email address."
```

aalto/zulip.rst
```diff
@@ -0,0 +1,75 @@
...
+**Organization settings / Video chat provider**
+
+  * Set to ``None``
+  * The default provider (Jitsi) has not been evaluated or approved by Aalto
+  * Integration with Aalto Zoom may come later on
+
+
+**Organization permisisons / Invitation settings**
+
+We cannot allow people to register freely using random emails, therefore:
```
rkdarst, line 35: Is this because of Aalto policies saying that we can't/aren't allowed to provide services to random people?  If so, we could say that so it's clear why.

aalto/zulip.rst
```diff
@@ -0,0 +1,75 @@
...
+   There are also `Rocket.Chat <https://rocket.chat/>`_ instances available, provided by ITS. Rocket.Chat instances can be requested from ``servicedesk@aalto.fi``
+
+Configuring your organization
+------------------------------------
+
+Below are listed the most important settings found under *Manage organization* in Zulip. There is no easy way for us to enforce these, so it is your responsibility as organization owner or admin to make sure they are set correctly. Make sure any owners/admins you appoint are aware of these as well.
+
+.. note::
+
+    Settings that are not mentioned here, you can configure to your liking. However you should still exercise care.
```
rkdarst, line 23:
```suggestion
    Settings that are not mentioned here, you can configure to your liking. However you should still exercise care, since you are responsible for the service and safety of your user's data.  If you would like advice, please ask us.
```

rkdarst (approved):
This looks good.  I added some suggestions below, and here are some other questions I can think of, as an organization owner:
- What is the lifetime of each course?  Can it last forever?  Should it?
- What are allowed use cases?  Can I use this as my research group?
- Link to example organizations, for example you could use scicomp.aalto.fi as a demo (?)

aalto/zulip.rst
piiroim1: Not an Aalto policy, but common sense. Registering with just an email address is an open invitation for bots etc.

aalto/zulip.rst
```diff
@@ -0,0 +1,79 @@
...
+**Organization settings / Video chat provider**
+
+  * Set to ``None``
+  * The default provider (Jitsi) has not been evaluated or approved by Aalto
+  * Integration with Aalto Zoom may come later on
+
+
+**Organization permisisons / Invitation settings**
+
+We cannot allow people to register freely using random emails, therefore:
```
piiroim1, line 39:
```suggestion
Do not allow registering with just an email without invitation. Doing so might cause bots and random people to join in your chat. If the chat needs to include external users, make it invite only.
```

rkdarst:
thanks!

PR #235, opened 2020-11-30
Update the zulip page

piiroim1:
* Add mention about chat instance lifespan and Spring 2021 semester.
* Add mentions about usage in grading and for internal/confidential
data.


rkdarst:
Sorry for the delay, thanks for this.

PR #245, opened 2020-12-20
Add a blog functionality

rkdarst:
- This uses ablog (https://ablog.readthedocs.io/), which seems to work
  OK but I think it is worth thinking about long-term maintanance that
  this adds to this site.  Should the blog be kept separate?
- There is a `blog/new.sh` script to automate making a new post.
- There are several test posts, which should be edited or removed
  before posting.
- I haven't done too much to try to integrate this into the main site:
  nothing on sidebar (is that wanted), the list on the main page could
  be adjusted some.
- Could this replace the existing "news" setup?
- Review:
  - Discuss questions above
  - Don't merge until there is some further discussion

### DIFF
--- a/blog/2020/12/hello-world.rst
+++ b/blog/2020/12/hello-world.rst
@@ -1,0 +1,13 @@
+:blogpost: true
+:date: 2020-12-19
+:author: Richard Darst, Test User
+:category: meta, blog
+
+..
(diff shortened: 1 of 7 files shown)


rkdarst:
We decided to do this separately.  Closing.

PR #251, opened 2021-01-04
Add "Intro to Linux and computing" course page

rkdarst:
- Draft, at everything is still to be discussed.  Even dates could be
  changed.
- Please review for both semantics and syntax.

### DIFF
--- a/training/index.rst
+++ b/training/index.rst
@@ -26,6 +26,7 @@ and/or Python, Matlab, HTCondor and many others.
    :maxdepth: 1
 
    scip/winter-kickstart
+   scip/intro-linux-aalto-computing
    scip/gpu-computing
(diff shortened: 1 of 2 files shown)


wmvanvliet:
Under software: how to obtain a bash shell on windows? One option is to SSH into an Aalto server, but then maybe some instructions are needed. Another option is to install WSL locally.

rkdarst:
For that, we have some sort of information for Triton, which can be
reused (and perhaps made standalone somehow)?

I was also going to propose binder for this: simple and effective,
though not so useful to the researcher useful long-term (for purposes
of a shell).

eglerean:
I merge pull request, and we can update BASH for windows later

PR #253, opened 2021-01-07
Added link to CR BASH installation page

eglerean:
Added link to CR BASH installation page


wmvanvliet:
I'm not a fan of using Git Bash. WSL is a much cleaner solution if you have a modern windows 10 installation. Of course, we cannot assume everyone will have the latest and greatest windows. For the windows XP and windows 7 holdouts, Git Bash is probably a good solution. I'll make some edits to this PR, hang on.

rkdarst:
This is good, but also we can recommend a SSH client installation, to use via shell servers (since this will be needed otherwise).  This starts to give a two-way option at the start, but options are confusing of course!

wmvanvliet:
Good point. Perhaps it's best to let Windows users just install PuTTY and SSH into a linux server.

rkdarst:
@wmvanvliet, while editing you can add a pure Binder possibility as an option: https://mybinder.org/v2/gh/AaltoSciComp/bash-binder/HEAD?urlpath=terminals%2F1  (repo: https://github.com/AaltoSciComp/bash-binder/)

For ssh, our standard instructions are here: https://scicomp.aalto.fi/triton/tut/connecting/#connecting-via-ssh
but that is in the context of Triton so I wonder if we should split it off somehow, to go into more detail.

We can also make a new "ssh" page under CodeRefinery installation instructions, and use that as standard place to link people.  We could even make a "shell" page that gives shell instructions, not focused on the git part.

Remember we can't even assume people with Aalto laptops can use WSL, if they aren't admin.

eglerean:
I agree that we should just recommend one option (local linux OR remote linux), or it's going to be extra work also for us. But maybe we should not assume that everyone can connect to a server..? I am thinking pedagogically: linux shell works with my files on my laptop (even though it is not running Linux OS) and eventually I can type command ssh servername and continue remotely

wmvanvliet:
IMO: the simplest is to suggest one SSH client (PuTTY) and let them connect to one server (taltta?).
Do all participants have an aalto account for sure?

wmvanvliet:
Binder is also super easy. But also very slow and generally not a good advertisement of the "power of the CLI".

wmvanvliet:
Ah:

> Aalto or university account useful but not needed

there goes the SSH plan.

wmvanvliet:
On binder:

> bash: nano: command not found

Also not great...

rkdarst:
I don't think we should assume everyone has Aalto account.  People from other places may watch, and we want to stream to the public.  My vision right now is something like:

- Mac, Linux: just start terminal!
- MyBinder LINK: this works for everyone
- Windows: you can install bash locally, see LINK for options.
- You can also connect to one of your university's servers via SSH.  This is especially needed if you are also joining the HPC kickstart, because it will be needed.  See this link for details: LINK .  If you are at Aalto, connect to server kosh.aalto.fi, other locations usually have their own servers.

rkdarst:
I'm working on fixing the binder link.  nano is easy, harder is getting the manual pages in there.  But they are coming, I hope.

wmvanvliet:
How about this?

training/scip/intro-linux-aalto-computing.rst
```diff
@@ -103,8 +102,26 @@ Preparation
...
-* It could be useful to install a SSH client, but we have a backup
-  plan if you don't. `Here some options for installing BASH on your computer <https://coderefinery.github.io/installation/bash/>`__.
+* You will need the BASH shell
+
+  * Linux: Open the Terminal application and type: ``bash``
+  * MacOS: Open the Terminal application and type: ``bash``
+  * Windows:
+
+    * If you have an Aalto account, download `PuTTY <https://www.putty.org/>`__
+      and use it to connect to ``kosh.org.aalto.fi`` (see this `screenshot
```
rkdarst, line 112:
```suggestion
      and use it to connect to ``kosh.aalto.fi`` (see this `screenshot
```
Normally I say kosh.aalto.fi because it's shorter and easier to remember (though inconsistend with other computers...)

training/scip/intro-linux-aalto-computing.rst
```diff
@@ -103,8 +102,26 @@ Preparation
 
 **Software installation:**
 
-* It could be useful to install a SSH client, but we have a backup
-  plan if you don't. `Here some options for installing BASH on your computer <https://coderefinery.github.io/installation/bash/>`__.
+* You will need the BASH shell
+
+  * Linux: Open the Terminal application and type: ``bash``
+  * MacOS: Open the Terminal application and type: ``bash``
```
rkdarst, line 108: Hm, is zsh the default on mac these days?  and on Aalot zsh could be the default, too.

In practice, I think what we talk about should be the same on zsh and bash.  Does anyone know of a case that we need to make a distinction? (for the shell intro part, not the Triton part)

rkdarst (commented):
looks good!  some suggestions

rkdarst:
If there is not much more, I say merge and iterate.

rkdarst:
I got the binder image working with bash, git, and manpages.  Since
- ubuntu docker images disables manpages by default
- repo2docker doesn't have a way to install man pages
- repo2docker doesn't provide hooks to run things as root
- there was some bug in latest python images which meant `apt install` did not work (signature error) to install ubuntu packages!!
- perhaps several other small annoying things to figure out and work around

this was for more annoying than I had hoped... but it is extendable as much as we might like now.

PR #255, opened 2021-01-07
Update mne.rst

eglerean:
updating module load mne after discussion

### DIFF
--- a/triton/apps/mne.rst
+++ b/triton/apps/mne.rst
@@ -15,4 +15,6 @@ In the directory::
 
 you can find the relase notes, the manual, and some sample data.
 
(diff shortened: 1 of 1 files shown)


triton/apps/mne.rst
```diff
@@ -15,4 +15,6 @@ In the directory::
 
 you can find the relase notes, the manual, and some sample data.
 
+We do not recommend using the MNE command line tools, a more modern solution is to use the MNE-python suite.
```
wmvanvliet, line 18:
```suggestion
We do not recommend using the MNE command line tools, a more modern solution is to use the [MNE-Python suite](https://mne.tools/stable/overview/cookbook.html).
```

PR #258, opened 2021-01-14
triton: Rename .slrm to .sh

rkdarst:
- During the doc cleanup, we decided to make all of the Triton scripts
  consistent, by renaming to .sh (instead of .slrm).
- I forget the exact logic for why we chose this extension, but it
  made sense at the time.
- Review:
  - CI check
  - Do we still want to do this?


rkdarst:
This is ready.

tpfau (commented):
Looks good

tpfau (approved):

PR #265, opened 2021-01-25
Add sphinx-{lesson,tabs,copybutton,togglebutton} extensions

rkdarst:
- These allow better buttons and tabs for exercises (and other places)
- Review:
  - general check
  - do we want to preserve the ability to build without any extensions
    installed?  Right now you can, and it will just degrade the output
    somehow.  Is an immediate hard error better?

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -40,6 +40,10 @@ extensions = [
 # extensions only if they are importable:
 optional_modules = [
     'sphinx_rtd_theme_ext_color_contrast',
+    'sphinx-copybutton',
+    'sphinx-lesson.directives',
+    'sphinx-tabs',
+    'sphinx-togglebutton',
     ]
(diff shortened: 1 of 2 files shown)


simo-tuomisto:
I don't think it's a problem that the output is degraded for anyone that builds without extensions. `requirement.txt` contains requirements for the actual output and if one wants to build with that output, installing the requirements is not a big hassle. If one wants to build the site with bad output it is their call.

I updated the `conf.py` imports. The syntax was wrong (python does not allow for hyphens in module names). Tested the build with [this sample lesson](https://raw.githubusercontent.com/coderefinery/sphinx-lesson/master/content/sample-episode-rst.rst) and extensions worked.

rkdarst:
oops!  Thanks!  Let's do it...

PR #271, opened 2021-01-29
scicomp/shell: Add section on variables

rkdarst:
- Variables are actually a quite important part of basic shell usage,
  at least when following other examples it is useful to know $NAME
  syntax.  Quickly introduce it here.
- Review: worth a detailed read: is there anything that can be
  removed?  Anything missing?


rantahar (approved):
I think it's clear, there is no repetition from this file and I have nothing to add.

PR #280, opened 2021-02-04
scicomp/zen-of-scicomp: Revise links, etc

rkdarst:


bast:
I think it's good for the moment. Right now nothing to add.

PR #284, opened 2021-02-11
aalto/aaltodata: Mention that users must be removed from groups

rkdarst:
- Previously, it didn't mention that group membership remains after
  a possible employment contract ends.
- Review:
  - general read
  - Is there anywhere else this should be mentioned?  There is a lot
    of text and maybe the important parts are too hard to find.

### DIFF
--- a/aalto/aaltodata.rst
+++ b/aalto/aaltodata.rst
@@ -129,9 +129,11 @@ everyone in the group has access.
    the necessary information (see bottom of page).
 -  Each group has an owner, quota on filesystems, and some other
    metadata (see below).
--  Group membership is per-account, not tied to employment contracts.
-   If you want someone to lose access, you have to explicitly remove
-   them when done.
+-  Group membership is per-account, not tied to employment contracts
+   or HR group membership.
+   If you want someone to lose access to a group you manage, they have
+   to be explicitly removed by the same method they were added (asking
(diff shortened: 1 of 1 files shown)


aalto/aaltodata.rst
```diff
@@ -126,9 +126,12 @@ everyone in the group has access.
       there are fixed multi-year projects, they can also get a group.
 
 -  Groups are managed by IT staff. To request a group, mail us with
-   the necessary information (see next point).
+   the necessary information (see bottom of page).
 -  Each group has an owner, quota on filesystems, and some other
-   metadata (see below),
+   metadata (see below).
+-  Group membership is per-account, not tied to employment contracts.
+   If you want someone to lose access, you have to explicitly remove
```
eglerean, line 133: If you want someone to lose access to a group you manage, you have to explicitly ask the administrators to remove the account from your group (group owners at CS dept can do this themselves).

aalto/aaltodata.rst
```diff
@@ -126,9 +126,12 @@ everyone in the group has access.
...
+   them when done.
```
eglerean, line 134: I made it explicit that they need to request for a removal (unless at CS since there group owners can do it themselves)

eglerean:
Tiny comment and edit. Otherwise this is ready to be merged

PR #288, opened 2021-02-15
help/index: Add link to zulip for courses

rkdarst:
- To get more users for Aalto Zulip
- Review: does this belong here?


eglerean:
I think it belongs there. I am merging

PR #290, opened 2021-02-16
More links!

eglerean:
Adding more courses not organized by us. Adding registration links that were missing.

### DIFF
--- a/images/levels.svg
+++ b/images/levels.svg
@@ -1,0 +1,726 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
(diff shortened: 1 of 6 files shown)


rkdarst (approved):

rkdarst:
While doing this big revision, should hands-on scicomp be more mentioned and integrated somehow?  Like, 'here is a basic outline, our courses are ..."?

eglerean:
yes! Even with a picture of the HoSC roadmap embeded in the /training/ page for example and a link to the Aplus too maybe?

rkdarst:
Yeah, good idea.  Perhaps now the picture could be improved (I wasn't
very good at making it).  The story is good but by now it could be
revised to make it more clear and user-centric, too.

eglerean:
I have added the image and local build looks good. I merge and then we can revise

PR #293, opened 2021-02-26
Asking for help: add fourth question about "what do you need"

rkdarst:
- A complicated request may never be answered becaues it takes too
  long to think about the solution, where some initial advice or "this
  will take too long, we recommend you find another solution" would be
  fine and fast.

- Inside motive: find a way to reduce the number of issues which are
  never answered, or only answered after a very long time, because
  they take too long.
- Review:
  - Is four too many key questions for a summary?
  - Review of wording, in particular can it be made shorter?


eglerean:
I think this is good. In general I am not expecting that a users who asks for help has read this page, so I think these questions are also important for us to remember to set expectations immediately when a request is received.

PR #302, opened 2021-03-12
Add a story about Singularity+MPI (Serpent)

sjjamsa:
The basic things I learned are now reported here. Unfortunately the Serpent source code is closed, so I can't really make a nice tutorial kind of presentation. One could consider writing a separate Serpent page and just add a link.


rkdarst (approved):
I haven't tested but after reading, it seems reasonable.  should I go ahead and merge?

rkdarst:
I will just merge this, we can improve later if needed.

PR #310, opened 2021-04-07
index: Reduce the number of items on the sidebar

rkdarst:
- Reduce items on sidebar, limit links on front page to those that are
  important.  The remaining links are put under the about/ section
  toctree (which is where I thought they could be anyway...)
- Rename "About ASC" to "About" to reflect bigger picture.
- Review: is this a good idea?


sjjamsa:
All good to me.

PR #312, opened 2021-04-13
Switch all --mem units to M (instead of unspecified)

rkdarst:
- Once in a garage, someone was confused about the default units.  An
  easy way to solve this is to always specific them.
- Add `M` as the dafult unit to every --mem and --mem-per-cpu, if not
  already specified.
- Found by clever grepping.
- Review
  - Do we want to do this?


sjjamsa (approved):

eglerean:
I think this is good to merge, or was there anything else pending?

rkdarst:
I think so

PR #322, opened 2021-05-24
training/scip/summer-kickstart: Update schedule from HackMD

rkdarst:
- As currently written in planning
- Mainly I updated day 1, day 2-3 are roughly like before and needs
  more edits later.
- Review:

  - general check, incremental work but this makes it look more correct


eglerean:
I merge for now, but there are couple of warnings we can fix later

PR #323, opened 2021-05-28
Add Julia course to the training page

rantahar:
Just so that people see it. Materials are not available yet.

Review:
 - Is this the right place for it
     - In terms of page content
     - In terms of file structure
- General language check for the course page.


rkdarst:
I think this is good to go.  Page structure looks good for now, try copying from (for example) the kickstart course page for a structure for a course that needs to be more involved.

PR #330, opened 2021-06-02
adding summer warnings

eglerean:


help/garage.rst
```diff
@@ -2,6 +2,20 @@
 Scicomp garage
 ==============
 
+.. admonition:: Summer News!
+
+   During the summer we have reduced support. 
+   * From 21/June until 8/August the garage is available only on request.
```
rkdarst, line 8: Should we say "ask in chat for garage" and specify channel/topic?

rkdarst (commented):
I hope we can still have something in the summer, even if not guaranteed.  While not perfect, we can say keep checking and can have small things scattered here and there.  Plus, those of us who *are* working, garage is a great time to chat anyway (to me, that was the main point of starting a daily garage last summer - not for others, but for us).  If we combine SciComp and CSIT and possibly others, it can be a good social time.

rkdarst:
I updated it based on discussion, not so pessimistic about never being there.  We will have to keep things up to date as we go.  I hope that we can keep the garage as a social time when people are around..

 I guess I'll merge and we can iteratively improve?

PR #338, opened 2021-06-07
no need for difficult exercises here

eglerean:
Exercises left are now optional and more of a discussion. I have removed things that would confuse a beginner (seen in previous kickstarts). This page should be presented fast so that more time can be spent on modules which is strongly linked to this page.

### DIFF
--- a/triton/tut/applications.rst
+++ b/triton/tut/applications.rst
@@ -134,6 +134,18 @@ modules available on your cluster.
 3. (optional) Discuss among your group what software you need, if it's
    available, and how you might get it.
 
+4. (optional) Figure out how to use ``tensorflow`` (this is not a software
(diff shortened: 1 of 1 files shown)


rkdarst:
Yeah, I guess these advanced things were made to give you something to do, even if not very useful.

But I think some of these advanced things have the benefit of letting you do *something*.  For example, they shouldn't be done in a course but could possibly be useful when you want people to do something (anything) to turn in to get the credits.  That may be a better way to look at them.

So I propose moving them to the end and saying "optional, advanced", maybe with a text section before that makes it clear that you shouldn't do them?  In the course we should really emphasize that you should not try them in in the course itself.

eglerean:
I agree with the proposal

eglerean:
ok re-added back those exercises

PR #349, opened 2021-06-29
triton/apps/gurobi: Update Gurobi page with newer infrormation.

simo-tuomisto:
This PR adds information on Gurobi usage in Python and Julia.


rkdarst:
This looks good to me!  I can't verify things but it does match what I have seen in the past!  I can't believe we haven't made this before.

PR #378, opened 2021-09-05
rkdarst/zulip upgrade

rkdarst:
- python-for-scicomp course page with basics
- training/scip/python-for-scicomp: Correct times, add tz converter
- aalto/zulip/: Update based on new zulip


rkdarst:
OK, so the new `git-pr` usage of `gh` gave a wrong message here, somehow it included messages from other commits.  I will need to work on that...

PR #390, opened 2021-09-21
Adding sftp-mounting to storage tutorial

tpfau:
Adding another (potentially convenient) way of mounting triton filesyystems remotely using sftp and proxy jumps.


sjjamsa:
Hi! Two comments:
1. Maybe mention taltta with kosh on the first time 
2. I would emphasis that the jump through kosh/taltta is there to pass through the firewall

sjjamsa (approved):

PR #414, opened 2021-10-15
Embed hpc-examples as a submodule

rkdarst:
- Add the submodule at triton/hpc-examples.  This can co-exist with
  triton/examples for a while, until they are merged.

- Currently, hpc-examples is not well organized, via the sphinx site
  most of the examples can't be reached.  Doing so would seem to
  require that a lot of minimally-functional .rst pages are made to
  embed the code, which isn't great but would allow us to explain some
  more.  There are possibly other solutions that should be
  investigated.

- This could be merged now (minimally working, and improved later), or
  it could be left as a draft and we can figure some more things out
  before merging.

### DIFF
--- a/.github/workflows/checkwarnings.yaml
+++ b/.github/workflows/checkwarnings.yaml
@@ -15,6 +15,8 @@ jobs:
         python-version: [3.8]
     steps:
       - uses: actions/checkout@v2
+        with:
+          submodules: recursive
       # https://github.com/actions/setup-python
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v1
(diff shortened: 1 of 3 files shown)


rkdarst:
Yes, long term that would be my goal.  You could say this is "step
one", get this working, then in a second step migrate existing
examples over and then rename hpc-examples to examples.

My quick effort showed there would be more effort than expected,
mainly updating all the index files in hpc-examples to work with
sphinx (and *also* work standalone on github).  I think this should be
easily solveable with some time, though.

rkdarst:
this is so old it's obsolete.  Closing and we can re-make.

PR #423, opened 2021-11-03
Adding Info-Box for GPU jobs

tpfau:
Adding an admonition to stress, that if submitting a job using srun (in a batch script) you need to re-request the GPU ressource.

### DIFF
--- a/triton/usage/gpu.rst
+++ b/triton/usage/gpu.rst
@@ -79,6 +79,14 @@ only difference is that they have nvidia kernel modules for Tesla cards.
 Running a GPU job in serial
 ---------------------------
 
+.. admonition:: Important note
+ 
+   When running a job using GPUs on the queue, be aware that you need to re-specify 
+   `` --gres=gpu:N`` in your function call if you run it via ``srun``. Otherwise the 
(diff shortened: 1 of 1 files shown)


triton/usage/gpu.rst
```diff
@@ -79,6 +79,14 @@ only difference is that they have nvidia kernel modules for Tesla cards.
 Running a GPU job in serial
 ---------------------------
 
+.. admonition:: Important note
+ 
+   When running a job using GPUs on the queue, be aware that you need to re-specify 
+   `` --gres=gpu:N`` in your function call if you run it via ``srun``. Otherwise the
```
rkdarst, line 85:
```suggestion
   Update 2021-09: We recommond that you do **not** use `srun` within the job script
   for GPU jobs, since something has changed in the recent Slurm.  We are working
   on a permanent recommendation/solution.

   When running a job using GPUs on the queue, be aware that you need to re-specify 
   `` --gres=gpu:N`` in your function call if you run it via ``srun``. Otherwise the 
```

tpfau: Sounds good,

PR #429, opened 2021-12-09
Quick start 

tpfau:
Adding a quick-start guide to triton.
The guide contains:
1. Quick setup of Linux/Windows machines to easily connect to triton using ssh keys.
2. Basic information about data handling on triton (data transfer)
3. Basic info on job submission 
4. Links to basic usage examples


rkdarst (commented):
OK, I've reviewed this.  I think rather than slow down we should merge and then continue to revise from there, big changes can take forever to make things "perfect"

After thinking more, I like the idea of:
- a quickstart page that sort of goes along with the "quick reference" that isn't about topics but "if you want to get something done, what do you need to do?"  I think that's the main point here.
- But some of the other "how to connect" and so on should go into improved tutorials or quick reference, and then be linked to there.  The tutorials should be improved to take in this spirit, make it more to the point of getting stuff done (at least easier to find that part!)

I will do some pushes to you branch for other edits/improvements, and then merge.  Might take a few more days though, not sure.

rkdarst:
I think this is close to ready (if not already)

scicomp/rsynconwindows.rst
```diff
@@ -0,0 +1,20 @@
+===========================
+Installing rsync on Windows
+===========================
```
rkdarst, line 3: So I wonder how practical it is to recommend this (even if it is the theoretical best method).  I remember the day of trying to help people transfer stuff by scp/rsync before the age of SMB mounts.  This wasn't fun and I'm not sure how many people even were able to do it.  That's when I realized we needed SMB mounting to make things reasonably usable for most people.

Considering this, do you have any more thoughts on the role of this and is there some other thing to recommend as the top-level option?  (see also other note which I think is in the comment feed, but not here anymore because the file moved).

rkdarst:
I will merge this now so it doesn't linger forever and I can continue with other improvements.  But, I would like to consider all of the points above at some point in the future.

PR #432, opened 2021-12-14
Slurm to sh

tpfau:
Additional changes to #258


rkdarst:
Oops, I just added these changes to #258 (I didn't see this).  I wasn't sure if doing something in master would conflict or not so did it there.  Let's see if there are any differences now...

tpfau:
That's perfectly fine. 
I just thought I could fix the mistakes I did.

PR #436, opened 2021-12-16
Adding collection information commands

riccardofelluga:
There are this two useful commands to list and get the details on the users' saved collections and they should be also listed in this table.


rkdarst:
Thanks!  Good info, I hardly ever use collections so wasn't sure what was useful.

PR #440, opened 2021-12-29
triton/ref: Add remotedata quick reference (transferrs and mounts)

rkdarst:
- Triton rsync and sftp transfers
- Triton mount paths
- Review: someone should read the text

### DIFF
--- a/triton/ref/index.rst
+++ b/triton/ref/index.rst
@@ -40,6 +40,11 @@ Storage
 
 .. include:: storage.rst
 
+Remote data access
+==================
+
+.. include:: remotedata.rst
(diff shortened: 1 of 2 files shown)


tpfau:
Would it make sense to add this on the data page of the quickstart guide? Or replace some part of the data page in the quickstart guide by it?

rkdarst:
I think I was inspired to make this based on what I saw in the quickstart page (it pointed out important stuff that was missing, plus I copied this).  So yes, combining is probably a good idea, I just didn't get around to it yet/didn't want to change too many things at once.

PR #443, opened 2022-01-04
triton/tut/serial: advanced exercise to show job env variables

rkdarst:
- Ideas from Simo

- The idea is it might be good to show that these environment
  variables exist.  But it may not be worth taking time from the main
  flow.  Making it an advanced exercise means that the people who have
  time and mental energy can explore it.
- Review: text and "is this a good idea?"


tpfau (approved):
Looks good

PR #451, opened 2022-01-15
about/index: Link FCCI Tech (/tech/) under /about/

rkdarst:
- Before it was not findable without knowing the URL.  Should we make
  it findable?
- Review: think if there is any reason not todo this.


eglerean:
This was opened in January (!): is it obsolete or do we merge it as it is?

eglerean:
This is still open :)  .... merge? maybe with past tense sentences?

PR #452, opened 2022-01-17
training/scip/winter-kickstart: Add final topis to day 3

rkdarst:
- As discussed in the meeting today
- Also add in "real-life examples of different languages", which we
  did one year and turned out well.
- Review: see if these are good descriptions or anything else makes
  sense.


eglerean:
Looks good to me. I merge for now. I guess questions about csc services can happen after 15:30

PR #456, opened 2022-01-19
training/scip/winter-kickstart: Schedule update

simo-tuomisto:
This PR changes schedules around and differentiates different topic blocks


rkdarst:
Looks good to me, as we discussed!

PR #458, opened 2022-01-19
add my name and link to presentation slides

bast:


rkdarst:
Thanks!  Good to have the first non-Aalto name added, hopefully it will lead to more.

PR #473, opened 2022-02-28
Update connection instructions for different OS, to reflect the OS 

tpfau:
Updated the remote access pages with tabs for each OS, containing the actual info for that OS.


rkdarst:
Great improvement, very much needed!

PR #475, opened 2022-03-03
training/scip/index: add links to old courses

rkdarst:
- Proposal: add links to the old courses.  While most of these pages
  get updated so no longer are active for the year in question, I seem
  to be always looking for these pages, and they aren't linked.
  - This is probably good for search engine visibility
- Problem: they get updated for later years.  But still, perhaps most
  people looking for what they could expect if they took the course
  again (and it provide a starting point for looking into the git
  history)
- I think this is more useful than a problem
- Review: decide if we should do tihs.


eglerean:
Looks great! I will merge for now, my idea is to get back the original pages and in the future create pages that are unique, e.g. https://scicomp.aalto.fi/training/scip/data-anonymization-workshop-2022/ vs https://scicomp.aalto.fi/training/scip/data-anonymization-workshop/ I

rkdarst:
Yeah, at first I liked the idea of "always updating the same page" but it's looking like we do value going back.

Or what if we linked to the old versions in Github history - it won't render nicely, but will work for some cases.  For example in a "see also" section at the bottom?

PR #479, opened 2022-03-10
Remove Marijn and Essi, Add Thomas to /about

rantahar:
@tpfau, can you check your entry and suggest specialties?

### DIFF
--- a/about/index.rst
+++ b/about/index.rst
@@ -180,7 +180,7 @@ better.*
 
    * * Thomas Pfau
      * Science-IT, Aalto RSE
-     * Software development, 
+     * Software development, Matlab, Linear/mixed integer programming, Constraint based metabolic modelling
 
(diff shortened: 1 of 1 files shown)


about/index.rst
```diff
@@ -178,13 +178,9 @@ better.*
...
 
-   * * Marijn van Vliet
+   * * Thomas Pfau
      * Science-IT, Aalto RSE
-     * Software development, data, neuroimaging
-
-   * * Essi Heikkinen
-     * Science-IT
-     * Natural language processing, teaching, documentation
+     * Software development,
```
tpfau, line 183: Things I would suggest:
, Matlab, Linear/mixed integer programming, Constraint based metabolic modelling

rantahar:
Just asking for a quick check before merging (or feel free to merge yourself)

PR #491, opened 2022-04-01
Adding Disk Info to HW overview

tpfau:
Adding a row with information about local disks to the Cluster Hardware Overview.
Admittedly, this table becomes large, but I have no idea how to best condense /split it to a form that is better fit for the general documentation layout.


rkdarst:
Seems about as good as it can be until we come up with some much different idea...

PR #497, opened 2022-04-07
Adding IdentityFile parameter for consistency

tpfau:
The current tutorial creates a non default name key, which needs to be indicated explicitly in the config. While this is mentioned before the Sample config file, it is missing in the config and I think the tutorial should be consistent, so that, if you follow all steps it works in the end.

### DIFF
--- a/scicomp/ssh.rst
+++ b/scicomp/ssh.rst
@@ -181,7 +181,7 @@ Now, you can just do ``ssh triton`` or ``rsync
 triton:/m/cs/scratch/some_file .`` directly, by using the ``triton``
 alias.  Note that the Triton rule uses the name ``kosh`` which is
 defined in the first part of the file. The ``IdentityFile`` parameter is 
-necessary only if you have a none default key name (like the one indicated).
+necessary only if you have a non-default key name (like the one indicated).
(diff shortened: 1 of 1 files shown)


scicomp/ssh.rst
```diff
@@ -167,17 +167,21 @@ This is placed in ``~/.ssh/config``::
...
+        ProxyJump kosh
+        IdentityFile id_rsa_triton
+
 
 Now, you can just do ``ssh triton`` or ``rsync
 triton:/m/cs/scratch/some_file .`` directly, by using the ``triton``
 alias.  Note that the Triton rule uses the name ``kosh`` which is
-defined in the first part of the file.
+defined in the first part of the file. The ``IdentityFile`` parameter is 
+necessary only if you have a none default key name (like the one indicated).
```
Gehock, line 184:
```suggestion
necessary only if you have a non-default key name (like the one indicated).
```

rkdarst:
Yeah, so like I said in the other PR, I am not sure if I would have a separate key file.  But it's there so we should add this to keep it consistent.  We should remember to update if we adjust this later.

PR #503, opened 2022-04-26
Ssh keys

tpfau:
Added OS specific instructions for ssh key generation and setup.
I tested the instructions on aalto-linux and windows. For the latter, I could not test the ssh-add part  since I don't have a windows machine with Admin privileges, and that's necessary for the Openssh-agent. I will give it a shot at my private machine at home.
Windows stores the .ssh directory in the user profile. Interestingly, ssh does decipher an key location indicated as `~/.ssh/key_name` even on windows, so the final configuration file examples don't need to be changed.

### DIFF
--- a/scicomp/ssh.rst
+++ b/scicomp/ssh.rst
@@ -41,10 +41,27 @@ Known servers
 
 You will not receive an authenticity prompt upon first login if the server's *public key* can be found in a list of known hosts. To check whether a server, *Kosh* for example, is known
 
-::
+.. tabs::
 
-    ssh-keygen -f /etc/ssh/ssh_known_hosts -F kosh.aalto.fi
-    ssh-keygen -f ~/.ssh/known_hosts -F kosh.aalto.fi
+  .. group-tab:: Windows
+          
+     ::
+      
+         ssh-keygen -f %USERPROFILE%\.ssh\known_hosts -F kosh.aalto.fi
+        
+  .. group-tab:: Linux
+     
+     ::
+      
+         ssh-keygen -f ~/.ssh/known_hosts -F kosh.aalto.fi
+
+        
+  .. group-tab:: Mac
+  
+     ::
+      
+         ssh-keygen -f ~/.ssh/known_hosts -F kosh.aalto.fi
+
 
 
 SSH keys: better than just passwords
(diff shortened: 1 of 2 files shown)


tpfau:
Needs some more fixes.

tpfau:
Please don't merge yet until the ssh-add command is tested.

tpfau:
Tested on Windows and works. So can be merged after review.

scicomp/ssh.rst
```diff
@@ -64,9 +81,28 @@ While there are many options for the key generation program ``ssh-keygen``, here
...
 Here are our recommended input options for key generation.
 
-::
+.. tabs::
+
+  .. group-tab:: Windows
+    
+     ::
+      
+         ssh-keygen -t rsa -b 4096 -f %USERPROFILE%\.ssh\id_rsa_triton -C "triton key for %USERNAME%"
```
rkdarst, line 90: Normally I would recommend one key per client computer, for all hosts it would connect to.  What are your thoughts on that?

I can think of some reasons for multiple (e.g. on Triton I have one passwordless one for internal ssh and one with password for git).  And maybe being more controlled has less risk of being interefred by other things.

Also, in the comments, I would always include the source computer.  When you look at Triton and see all the old keys there, you want to be able to easily see what is still relevant and what is obsolete.  (at least for me default sometimes seems OK)

scicomp/ssh.rst
```diff
@@ -75,21 +111,79 @@ Copy public key to server
...
 In order to use your key-pair to login to Triton, you first need to securely copy the desired *public key* to the machine with ``ssh-copy-id``. The script will also add the key to the ``~/.ssh/authorized_keys`` file on the server. You will be prompted to enter your Aalto password to initiate the secure copy of the file to Triton.
 
-::
-
-    ssh-copy-id -i ~/.ssh/id_rsa_triton.pub login_name@triton.aalto.fi
+.. note:: Connecting from outside of the aalto network
+   
+   The following command works, if you are within the Aalto network or if you are connected via vpn. If you log in from outside Aalto, 
+   you first need to set up the key for the login server (e.g. kosh). This can be done by replacing ``login_name@triton.aalto.fi`` by  ``login_name@kosh.aalto.fi`` in the following command.
+   If you connect from outside aalto it is usefule to first set up the config file :ref:`as described below<example_config_for_ssh>`.
```
rkdarst, line 118:
```suggestion
   If you connect from outside aalto it is useful to first set up the config file :ref:`as described below <example_config_for_ssh>`. 
```

scicomp/ssh.rst
```diff
@@ -75,21 +111,79 @@ Copy public key to server
...
+    	 You will need administrative permissions to be able to start a ssh-agent on your machine that can store and handle passwords. 
+    	     
+         1. Open *Services* from the start menu
+
+         2. Scroll down to *OpenSSH Authentication Agent* > *double click*
+
+         3. Change the *Startup type* to *Automatic (Delayed Start)*,
+            or anything that is not *Disabled*
+    
+         4. ``ssh-add %USERPROFILE%\.ssh\id_rsa_triton``
```
rkdarst, line 166: With the default key name, is it automatically found?

scicomp/ssh.rst
```diff
@@ -75,21 +111,79 @@ Copy public key to server
...
+         
+     
+    
+  .. group-tab:: Linux
+     
+     ::
+     
+         ssh-add ~/.ssh/id_rsa_triton
+        
+     If you are unsure whether a ``ssh-agent`` process is running on your machine, ``ps -C ssh-agent`` will tell you if there is. To start a new agent, use ``eval $(ssh-agent)``.
```
rkdarst, line 176: do Aalto ubuntus use some sort of Ubuntu-managed thing different from ssh-agent, or is that ssh-agent?  Hm, I should check myself...

rkdarst (approved):
This was very needed!

I have some comments below (at least one for me to check).  It's really interesting to think about this, since there are so many different things that go into it and are needed to make things simple but without things that will catch some people.

scicomp/ssh.rst
tpfau: I never tried this with ssh-add. Anyone else?

scicomp/ssh.rst
tpfau: They use ssh-agent, but I'm not sure, whether you can start it yourself.

scicomp/ssh.rst
rkdarst: ssh-add with no arguments uses the default key location.  I only give a key name where I have a separate key with password, separate from the default passwordless one.

PR #505, opened 2022-05-10
Not all people, just all Aalto people.

sjjamsa:


help/garage.rst
```diff
@@ -43,7 +43,7 @@ What kind of issues can we help with:
 
 Notes:
 
-* All garages are open to all people.
+* All garages are open to all researchers working in Aalto.
```
tpfau, line 46: Are we open for student questions as well?

eglerean: Short answer is no. I think the consensus was that in some cases they can be researchers also when they are doing course work closer to research (e.g. special assignments under PI/postdoc supervision). In general we won't help them with their homework :)  but I would say it is kind to direct them towards the right path (e.g. we can't help you with your python code, but here a course you can take by yourself)

eglerean: And often if a student has a triton account, then it is very likely doing some research or special assignment under research group supervision

rkdarst, line 46:
```suggestion
* All garages are designed for researchers and staff working in Aalto (or those who have a need to contact us).
```

"open to" → "designed for" to not necessarily exclude others (e.g. people who come specifically to talk to us)
"and staff": maybe this makes it too broad and should be removed
"or those who have a need to contact us": e.g. the CSC people or other service teams.  Or is this obvious and should be removed?

I do agree with the sentiment of more clearly saying who it is for (though I don't mind having occasionally other people dropping by, at our level it's nice to hear what happens other places).

rkdarst (approved):

help/garage.rst
rantahar: For the last part, I don't think it's too obvious. I would be anxious about asking those questions if I thought the event was meant for something else.

We occationally have people asking legitimate questions who think the garage might not be the right place because it's not about one of the focus days or not today's focus or...

PR #509, opened 2022-05-23
tech/user-support-responses: Based on request, "how to answer user requests"

rkdarst:
- My notes from my years of experience + observing people, written by
  request of someone.
- Unsure if this is obvious or not, but we can start collecting ideas
  somewhere, I guess.
- Review:
  - general reading, any further ideas to add?
  - This is incomplete so far

### DIFF
--- a/tech/index.rst
+++ b/tech/index.rst
@@ -16,6 +16,10 @@ explanations.
 We share what our practices are, what we have learned, and informally
 discuss.
 
+.. seealso::
+
+   :doc:`user-support`
+
 
 
 Practicalities
@@ -138,6 +142,10 @@ Topics 2021
 - Richard Darst
 
   - Online courses and CodeRefinery
+
+- Respectfully and efficiently handling user support requests
+
+  -:doc:`user-support-responses`
 
 - Science-IT data management: policies and procedures
 
--- a/tech/user-support-responses.rst
+++ b/tech/user-support-responses.rst
@@ -1,0 +1,184 @@
+How to actually respond to user support requests?
+-------------------------------------------------
+
+I've been asked before, "how do you actually respond to customer
+support requests?".  There are some obvious answers (be polite, try to
+answer, etc), but are there any specific references for research
+computing / scientific computing support staff?  This page collects my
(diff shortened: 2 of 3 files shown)


bast:
- document what you know and learn as you are working on the issue even if it is not solved yet
  - keeps requestor in the loop who is happy to see progress
  - makes it easier for the next staff who might need to take over the ticket to catch up

bast:
- even if it takes a while to find the right staff to answer, avoid long silence towards requestor who might not know that the ticket is being passed around

bast:
- communicate utmost respect. do not leave any room for misunderstanding. avoid jokes, avoid sarcasm. even phrases like "good luck" in the wrong context can be perceived as offensive.
- avoid too short answers only pasting a documentation link and "let me google that for you" answers

johanhellsvik:
- To the extent reasonable, strive to keep separate issues X and Y in different tickets / email threads. This makes it easier to keep track on what issues have been fully understood / answered / resolved; and what issues are open.
- In addition to communication with the user, add any relevant comment(s) for yourselves and your colleagues in the ticket. This can be for instance thoughts and suggestions on which you are not sure. These can be better to share only with your colleagues rather than to pass too many open ended points to the user.

rkdarst:
If not already said:
* Answering at too low a level can be condescending.  But assuming the user knows more than they do can create massive chaos, completely frustrating everyone and destroying self-confidence.  (happened to me when trying to help someone debug an off-grid solar-powered electrical system at a cabin.  I assumed too much and probably didn't make the person feel good about the questions.)
  * Benefit of unknown questions getting a first live chat (e.g. garage, see flowchart here: https://scicomp.aalto.fi/help/

rkdarst:
Thanks for the comment!  I have finally incorporated them.

This wasn't designed to be a general "how to use ticketing systems" type of thing, but instead focused more on the respect side of things.  Still, I added all the above ideas and probably will expand it more with a broader set of ideas of how to use the tools well.

I'm going to call this ready for merging - then we can keep iterating.  Maybe a big iteration will come after I give a proper presentation on it - perhaps we can arrange that together.

PR #511, opened 2022-06-01
training/scip/kickstart-2022-summer: Proposal for schedule update

rkdarst:
- Round out the schedule some more, make it look more interesting.
- Remove SciComp workflows and tools of SciComp (put under video lists)
- Add "about us", "full example workflows", "cooking metaphors to
  parallelism", and "environments in Python and R", all 25 minutes.
  Unsure if these are the best options but looking at old chats I see
  them mentioned and they seem like good options.
- Review: analyze, maybe quick merge and iterate.


rkdarst:
merging to iterate

PR #516, opened 2022-06-06
triton/apps/python-conda: New page with information on using conda

simo-tuomisto:
This PR creates a new page: Python environments with Conda.

Contents include:
- Adding a new page on conda environments.
- Sharing info-box on conda activate vs source activate between the new page and the old python page.
- Adding new example environments for tensorflow, pytorch and generic cuda environment.


triton/apps/python-conda.rst
```diff
@@ -0,0 +1,461 @@
...
+
+Below we have a simple ``environment.yml``:
+
+.. code-block:: yaml
+
+  name: conda-example
+  channels:
+    - conda-forge
+  dependencies:
+    - blas * mkl
```
rkdarst, line 54: blas seems like something beyond what most peolpe would use.  what about an example of pinning a ML library to a specific version as the starting example?

rkdarst (approved):
Looks good, one small comment but of course we can keep improving...

triton/apps/python-conda.rst
simo-tuomisto: @tpfau had the same comment. Fixed it in the new commit. MKL & BLAS is mentioned later in the complex environment section.

simo-tuomisto:
This should be good to go.

tpfau (approved):
Looks Good

PR #517, opened 2022-06-08
Change metavariables no N uniformly, instead of <n> and $n

rkdarst:
- After discussion about what the best meta-variables syntax is, this
  is what I thought was the least bad.  Not perfect, but uniform and
  easy to interpert for new users with a one-to-one character
  replacement with what they should do.
- Review: confirm this is something we want to do.


simo-tuomisto:
Looks good to me.

PR #534, opened 2022-09-07
training/scip/python-for-scicomp-2022: template copied from last year

rkdarst:
- Merge and improve, maybe after we know dates


eglerean:
sorry for having missed this. It seems that the page is basically the same I added (minus the dates and the registration) so maybe we can close without merging?

PR #537, opened 2022-09-12
Updates for fall 2022 teaching

eglerean:


rkdarst:
Looks good.  Failure is unrelated, I will fix it in another push.

I made another PR with python-for-scicomp-2022, but it has no real content changes, so I propose we merge this instead.

eglerean:
Great, I missed the other PR I guess. This here has dates and the registration form open. The error is in the workflow page, pointing to wrong help page?

rkdarst:
I think the link should be fixed now, even though re-running didn't run on my new push.  I propose merge and then fix if it's still broken.

PR #539, opened 2022-09-16
Adding OVERRIDE to doc

tpfau:
Seems like there are several versions of e..g. tensorflow now, where simply specifying the cuda version is insufficient and you need to actively override the conda CUDA detection mechanism. 
Added this to the docs.

### DIFF
--- a/triton/apps/python-conda.rst
+++ b/triton/apps/python-conda.rst
@@ -265,6 +265,7 @@ reproducible.
 Creating more complex environments
 **********************************
 
+.. include:: /triton/examples/cuda/cuda_override_hint.rst
 
 Creating an environment with CUDA toolkit
 -----------------------------------------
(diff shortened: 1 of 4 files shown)


simo-tuomisto:
What is the use case where one would need to override the CUDA version? Installing tensorflow or pytorch with incompatible CUDA version sounds like a risky proposition.

tpfau:
> I looked into this a bit and was surprised that this autodetection system is suddenly a problem. Previously conda & mamba would happily create an environment even when the runtime `__cuda`-requirement was missing, which made installations a lot smoother. Something has obviously changed somewhere upstream.
That's what I noticed too, otherwise I wouldn't have made this explicit. Interestingly, even the OVERRIDE environment variable seems to have changed at some point.

PR #548, opened 2022-10-26
training/scip/python-for-scicomp-2022: Clarify that anyone can register

rkdarst:


teemu-ruokolainen:
Looks good, useful addition.

PR #553, opened 2022-11-11
triton/tut/connecting: Update Triton+Windows instructions

rkdarst:
- It didnt' say how to start PowerShell, and I had to guess.
- Also, can someone add specific informations about how to start the
  WSL terminal?

### DIFF
--- a/triton/tut/connecting.rst
+++ b/triton/tut/connecting.rst
@@ -113,10 +113,10 @@ worth making yourself comfortable with this.
 
    .. tab:: Windows with PowerShell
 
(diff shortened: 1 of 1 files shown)


tpfau:
One thing I noticed:
At least for connecting and having a shell, I would personally nowadays suggest users to just use the OOD triton shell access. The huge benefit of it is that you don't need to set up anything locally, and might actually feel more "natural" to a windows system.

tpfau (approved):

tpfau (approved):
Sorry, was testing if I can withdraw an approval.

PR #561, opened 2022-11-24
Add Mastodon feed to the sidebar

rkdarst:
- Could this replace our news thing?  I don't think it's ready but it
  shows the idea.
- Problem: I don't like the idea of embedding other websites, which
  causes third-party requests.
- Problem: It still needs some styling
- Problem: I'm not sure if having this on every page is good for
  Mastodon's health - does it cause too many requests?  Some parts do seem cached.
- Possibly better: find some other solution that adds the news here
  (as in, not a Mastodon feed but yaml-based news feed here - I could
  make a sphinx extension)


rkdarst:
I think this is good enough to go now

PR #569, opened 2022-12-30
fix typo: shh -> SSH

bast:


bast:
I am currently rewriting the Norwegian documentation about SSH inspired by Aalto's and will credit and link to Aalto's docs.

rkdarst:
I'm not at a computer now to approve, but maybe now is a good time to think about how to de-duplicate and share more. I still don't have any good ideas about how t ada

-------- Original Message --------

rkdarst:
thanks!

bast:
Indeed it would be good to figure out good ways to de-duplicate. This is what I came up with on the Norway side: https://documentation.sigma2.no/getting_started/ssh.html

Imagine we could join forces and create amazing documentation for all the centers.

tpfau:
In my opinion the most important bit for de-duplication is to "fragmentize" the docs. Essentially working a lot with includes. This way, you can have a lot of "general" information, but for examples you can use "localized" versions. (for us triton.aalto.fi, for you saga.sigma2.no. 
An alternative would be to use config files (i.e. the config says: server_address = triton.aalto.fi; server_abbreviation= triton, server_ssh_jump=kosh.aalto.fi ...) and the docs have these elements as placeholders? However I think this approach does require substantuially more initial work (and care). 
But what should be possible is to create a generic "ssh docs" page as a github repo that can be added as a submodule to existing docs and just needs to be linked/the missing example files need to be added locally. If we can set up a fixed structure like:
`/ssh/` for the ssh submodule, and `/local/ssh/` for all the "includes" that contain local data, we can essentially use the same source data for most of the pages.

rkdarst:
I still think that this extension (that I made...) is one of the best
ways to handle these differing versions.  Includes are good but I
think they might make it too hard to read if almost everything is
included:

https://github.com/NordicHPC/sphinx_ext_substitution

PR #571, opened 2023-01-09
_templates/layout: Add a way to request improvements to the page

rkdarst:
- Common problem: pages are old or out of date.  How can we lower the
  threshold for reporting this?
- This adds a link to create a new github issue, with some stuff pre-filled.
- Is this the best way to manage it?  Should there be a private way to
  submit issues (that doesn't need a github account) too?
- Someone who can do styling could take a look and see if it looks
  good enough.

### DIFF
--- a/_templates/layout.html
+++ b/_templates/layout.html
@@ -4,8 +4,13 @@
 
 {{ super() }}
 
+<!-- Request improvements to this page -->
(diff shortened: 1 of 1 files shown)


rkdarst:
This is what it is now:
image

Is styling OK?  Link target and other stuff can be improved later.

rkdarst:
I am merging this so that I can iterate some more.

Issue #23, opened 2018-03-08
Tensorflow (and other GPU package) instructions need updating


rkdarst:
Tensorflow instructions need a lot of updating
- asking for too many resources
- cuda/cudnn not needed to be loaded
- just load anaconda

in general all the deep learning instructions should probably be updated!

PR #572, opened 2023-01-09
Conda/Python examples update

simo-tuomisto:
Lots of changes to Conda/Python examples:

- Change headers for CUDA toolkit, PyTorch and Tensorflow installation sections so that they appear in sidebar.
- Made CUDA override hint a hint that can be included in various different locations. Included it in Conda, PyTorch and Tensorflow pages.
- Small changes to CUDA override hint. Added references to Conda and Conda-Forge docs.
- Fixed naming of "CUDA" to be consistent with the actual name.
- Removed mentions to old software from pytorch / tensorflow pages.
- Removed line-end whitespaces.
- Fix Python page to include the correct OpenMP example

### DIFF
--- a/triton/apps/python-conda.rst
+++ b/triton/apps/python-conda.rst
@@ -109,7 +109,7 @@ Packages in ``environment.yml`` can have version constraints and version
 wildcards. One can also specify pip packages to install after conda-packages
 have been installed.
 
-For example, the following 
+For example, the following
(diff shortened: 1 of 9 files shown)


tpfau (approved):
Looks good.

We should start making a lot more use imports to replicate redundant information that is important at multiple places...

tpfau:
closes #23

PR #577, opened 2023-01-21
At the top of pages, list the last edited date

rkdarst:
- Previously, this was in the footer of each site.
- Now, instead of "Edit or Github", it says "Last edited %Y %b %d",
  and if the page is edited before a configurable threshold, it will
  put the date in `darkred`.
- This doesn't solve unmaintained pages, but it provides a more clear
  indication to users for what is up to date or not.  I think that
  keeping everything up to date is almost impossible, so instead let's
  make it easier for users to understand what is recent and not.
- Review: check preview and see.

### DIFF
--- a/_templates/breadcrumbs.html
+++ b/_templates/breadcrumbs.html
@@ -1,0 +1,14 @@
+{%- extends "!breadcrumbs.html" %}
+
+{% block breadcrumbs_aside %}
+<li class="wy-breadcrumbs-aside">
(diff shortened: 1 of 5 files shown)


rkdarst:
This is a preview of this pull request:
preview of this PR

It also adds `(Edit)` as a link in the sidebar.  The rest of that stuff was already there.
sidebar preview

rkdarst:
I want to use this as a demo this afternoon, so I will merge this and we can revert or improve later.
Unless there are comments in the next short time.

PR #578, opened 2023-01-25
rse: Preview interfaces are a high-impact service.

rkdarst:
- This is something that was hinted, but based on a talk I just saw
  the idea of public preview interfaces should be emphasized even
  more.  ChatGPT provides a good example that everyone knows about
  now.
- Review: what's the right terms to describe this, especiall for the
  index page?  I'm not quite happy with my descriptions.


rantahar (approved):

rantahar:
This is an important point and good to add. Also the phrasing sounds good to me, but of course I don't know exactly what you have in mind. 

Maybe merge and think about improving it later.

PR #584, opened 2023-02-07
scicomp/git-annex: Add mentions of DataLad

rkdarst:
- Based on mastodon threads.
- Review: check accuracy, check if the RDM-focus is justified, check
  if it should be even more strongly recommended than is written here.

### DIFF
--- a/scicomp/git-annex.rst
+++ b/scicomp/git-annex.rst
@@ -6,8 +6,12 @@ Git-annex for data management
    `Video intro to git-annex, from Research Software Hour <https://www.youtube.com/watch?v=NKCBXOfyoyM&list=PLpLblYHCzJAB6blBBa0O2BEYadVZV3JYf>`__.
 
    `DataLad <https://datalad.org/>`__ is a researcher/research data
-   management focused frontend to git-annex.  You might want to start
-   here instead (and then you get everything below).
+   management focused frontend to git-annex.  This page is a
+   relatively technical introduction to what goes on inside of
+   git-annex, so the `DataLad handbook
+   <http://handbook.datalad.org/en/latest/>`__ might be a better place
+   to start, and then consult this page for another view / more
+   detailed information.
 
 `git-annex <https://git-annex.branchable.com/>`__ is a extension to git
 which allows you to manage large files with git, without checking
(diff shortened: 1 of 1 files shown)


rkdarst:
@jcolomb - do you think this is accurate?  Should I be even more bold and tell everyone to start with datalad as an interface, and get all the things you see below in a more usable package (yet still usable independently)?

jcolomb:
may well depend on your readership, the strength of datalad compared to git annex directly is in (as far as I understand it):
- setting rules about what is annexed and what is not, then adding everything to git or to git annex via a single command.
- using submodules
- run command (get files,  do the analysis, drop them again + add metadata in git commit)

For most of my readership, datalad is still too complicated...

the GIN-cli is closer to git-annex, but works only for GIN repositories I think. (https://github.com/G-Node/gin-cli)

I would add a note at the beginning redirecting readers to these tools depending on their use-case.

BTW: for me it was super interesting, helped to understand the core functionalities, and will help me thinking about the tool. You may want to explain better what you meant with large files are put in chunks, I am not sure I understood.

scicomp/git-annex.rst
```diff
@@ -5,6 +5,9 @@ Git-annex for data management
```
jcolomb, line 5: This presentation is mostly meant for people with experience with git and the command line, who want to know more about git annex. If you feel overwhelmed by the content, you may want to check the [datalad handbook](http://handbook.datalad.org/en/latest/) for a smoother entry in the use of command line and git annex functionalities.
Also note that git annex solves the issue of big files, but not numerous files (git starts to slow down with several hundreds of files). To deal with numerous files, one can use git submodules, again datalad may be a tool of choice for implementing submodules.

jcolomb:
```suggestion
We hope you will enjoy reading this presentation of the main functionalities of git annex. It proves interesting to follow, as to understand the logic of the tool, even if you may move to using datalad later on.
```

jcolomb (commented):
putative intro, you may want to add a word about the quality

scicomp/git-annex.rst
rkdarst: I've added a lot of the above to the comment on the top (slightly rephrased but it was a good idea!)  I'm sorry for taking so long to review.

PR #591, opened 2023-02-21
Nbgrader JupyterLab

tuncbkose:
Added tutorial on submitting nbgrader assignments in JupyterLab, and references to the tutorial and change to JupyterLab UI.


tuncbkose:
Forgot to add a note on duplicating cells, will do it soon.

aalto/jupyterhub-instructors/news.rst
```diff
@@ -1,6 +1,13 @@
 jupyter.cs news
 ===============
 
+Spring 2023
+-----------
+
+* JupyterLab is now available and is the default
+  for new course servers. If you'd like to continue
+  using Jupyter Notebook for your courses, let us know
+  when requesting.
```
rkdarst, line 10:
```suggestion
  when requesting a new course.  JupyterLab now supports everything nbgrader needs,
  though the user interface is slightly different.
```

aalto/jupyterhub/nbgrader-jupyterlab.rst
```diff
@@ -0,0 +1,83 @@
...
+Go to https://jupyter.cs.aalto.fi and sign in with your account to access JupyterHub.
+You can find more general information about it in :doc:`../jupyterhub`.
+Select your course and start a session. Your will see that you are under the ``notebooks/`` folder,
+which is your home directory. Make sure you put everything important here, files saved
+elsewhere will get deleted when your session ends.
+
+.. figure:: img/home.png
+   :align: center
+|
+You can see currently available assignments by going to "Nbgrader->Assignment List"
```
rkdarst, line 21: If this is supposed to go in the figure caption (which I think it is, right), this is how it would be done.  I can fix the rest if this I correct:
```suggestion
.. figure:: img/home.png
   :align: center

   You can see currently available assignments by going to "Nbgrader->Assignment List"
```

aalto/jupyterhub/nbgrader-jupyterlab.rst
```diff
@@ -0,0 +1,83 @@
...
+.. figure:: img/downloaded.png
+   :align: center
+|
+Follow the instructions in the notebook, and don't forget to save your work!
+Once you are done, you can go back to the Assignment List and press the "Submit" button.
+You can submit an assignment multiple times, but your instructor will probably use
+only your latest submission.
+
+At this point, you can delete the assignment files from your ``notebooks/`` directory if you'd like,
+but it is often a good idea to keep a copy of your work somewhere.
```
rkdarst, line 50: It actually does cache a copy in `.local/share/jupyter` (or somewhere like that - I didn't check).  Should this be mentioned or is it needlessy confusing and detailed?  I don't think we mention it anywhere else, and the only major significance is if it gets too big and exceeds their quota...

rkdarst (approved):
This is very good stuff!  I'm sure we will keep iterating but there are some initial suggestions below, especially the caption thing.

aalto/jupyterhub/nbgrader-jupyterlab.rst
tuncbkose: In general, I didn't intend the text to be captions (they are not optional in going through the tutorial), and most of them refer to the images below them rather than above. But I can change them if you think it makes more sense/looks better.

aalto/jupyterhub/nbgrader-jupyterlab.rst
tuncbkose: The cache affects functionality: if you submit an assignment and then remove it (e.g. from `notebooks/`), you can still fetch the feedback for it. In fact, you need the cache to fetch feedback at any point.

So I think we can either be explicit about its existence and tell users to not mess with it unless they understand the implications, or continue having obscurity and not worry about someone doing something wrong.

In an ideal world, I think the latter is better because everything that would take up a lot of space should be in `coursedata/` instead, so a few notebooks shouldn't create problems. In practice, I don't know if that is the case.

PR #596, opened 2023-03-07
Update python.rst

teemu-ruokolainen:
Initial proposal, can be modified as necessary, no need to merge as is:

Remove mention of activating conda environment in slurm batch script with "source activate"

Add example of activating conda environment in slurm script.

edit: checks fail, I will check it out.


teemu-ruokolainen:
I wonder what the problem was with the checks, it seems they remained failed?

PR #600, opened 2023-03-22
Comsol focus day and other comsol updates

sjjamsa:
Updates to OOD documentation would be beneficial, as I'm linking it.


rkdarst:
The comsol focus days are a great idea!  Quick scan of the rest looks good, too.

PR #608, opened 2023-05-22
triton/tut/connecting: Improve tutorial

rkdarst:
- Slight arrangement at top (adding more to the abstract).
- Hide SSH config file in dropdown box, emphasize checking the SSH page
- Solutions to exercises
- Minor other updates
- Review: worth reading it

### DIFF
--- a/triton/tut/connecting.rst
+++ b/triton/tut/connecting.rst
@@ -21,16 +21,12 @@ for this series of tutorials).
 
 .. admonition:: Abstract
 
-   * You can connect to Triton via ssh
+   * The standard way of connecting is via ssh, but Open OnDemand and
+     Jupyter provide graphical environments that are useful for
+     interactive work.
    * Host name is ``triton.aalto.fi``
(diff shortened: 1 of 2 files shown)


triton/tut/connecting.rst
```diff
@@ -271,34 +278,59 @@ Exercises
    ``top`` works almost as well).  What else can you learn about the
    node?
 
+   .. solution::
+
+      The node was last rebooted 18 days ago, and the load average
```
Mirasalm, line 283: Maybe clarify the solution a bit by appending to the start something like: "You should see something like this. From this example output we can tell that..."

Mirasalm, line 283:
```suggestion
      You should see something like this. From this example output we can tell that the node was last rebooted 18 days ago, and the load average
```

Mirasalm (approved):
I suggested a potential rewording, but outside of that everything seems good.

PR #609, opened 2023-05-22
triton/tut/cluster-shell: Fill out tutorial

rkdarst:
- Now has exercises and more text
- Review: worth reading it


Mirasalm (approved):
Overall changes seem good. Only minor thing to think about is that the pi.py example has people running the script directly on login node. The load is probably negligible anyway, so it's just a question of if we should teach people the correct way from the get go, or if it's best to keep it like this for simplicity.

PR #611, opened 2023-05-28
triton/tut/applications: Revise for kickstart

rkdarst:
- Instead of tool-based approach, go over categories: stuff already
  installed, stuff we install, stuff you need to install yourself.
  This hopefully is nicer and more useful to read.
- Most important stuff is still there, but I reomved some stuff that
  seemed excessively verbose.
- Review: worth a read

### DIFF
--- a/triton/tut/applications.rst
+++ b/triton/tut/applications.rst
@@ -43,52 +43,74 @@ Linux.
    more common, but they are less standardized.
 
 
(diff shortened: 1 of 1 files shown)


Mirasalm (approved):
Good changes. In particular the new structure makes the page a lot easier to skim though if you are looking for something in particular

PR #613, opened 2023-05-29
triton/tut/serial: further update, minimize now that slurm page exists

rkdarst:
- General revisions, nothing too big
- Move away "partitions" now, add a shorter version to slurm
- Review: worth a read


Mirasalm (approved):
Good changes. Making the first batch script do something that actually gives you a relevant output file is a good idea. Also good reduction of repetition.

PR #616, opened 2023-05-30
Fix typos

pitmonticone:


rkdarst:
Thank you!  For some of these, I wonder how I wouldn't have seen them before... well, good you found them!

PR #619, opened 2023-05-31
updating the important green box with TL;DR

eglerean:
I turned the box into a TL;DR with relevant info. Please improve


Mirasalm:
All good information so merging it to get the information visible sooner than later. We can make adjustments afterwards if necessary.

PR #626, opened 2023-06-05
triton/usage: How big is my program

simo-tuomisto:
Adding a new page on how to calculate a program size.


triton/usage/program-size.rst
```diff
@@ -0,0 +1,290 @@
...
+- Physics simulation codes will usually integrate equations
+  in discrete time steps.
+- Markov chains do the same calculation for each node in the chain.
+- Deep learning training does training over multiple epochs and the
+  epochs themselves consist of multiple training steps.
+- Running the same program with different inputs.
+
+If this is the case, it is usually enough to measure the time
+taken by few iterations and from that information extrapolate the
+total runtime.
```
rkdarst, line 192: Subtracting startup time if significant?

triton/usage/program-size.rst
```diff
@@ -0,0 +1,290 @@
...
+
+Simple measuring stick: Your own computer
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you know nothing of your program, you can still probaly answer this question:
+
+**Does the program run on my own computer?**
+
+This can give you a good baseline on how big your program is.
+In general, you can use the following estimates to approximate your computer:
```
rkdarst, line 116: In or around 2022-2023?

rkdarst (commented):
This is really great.  Two minor inline comments.

What about some mention, before getting to the estimation strategies of something like: "It's OK to run it several times and see what happens while adjusting parametrs, without knowing exactly how much resources it needs.  Trial and error is often faster, but it's best if you combine that trial-and-error with the insights below."

PR #637, opened 2023-06-14
triton/tut/array: Comment on 2D array scanning

rkdarst:


sjjamsa:
Good :-)

Your code is pseudo-code, right? Let's say so explicitly, because that isn't something one could write in bash, is it?
Here is my proposal:

> Well, you can map 1D to 2D via the following pseudo-code:

PR #639, opened 2023-06-22
Formatting and highlighting on Python pages.

rkdarst:
- "console" highligher makes `$` prompt un-selectable
- while not done here, `.. highlight:: console` at the top makes
  default literal blocks highlighted that way.
- Review: automerge


Mirasalm:
The copy button on the side of code blocks still copies the dollar sign, even if it's not selectable otherwise. Is there a way to stop it from happening? Otherwise it somewhat defeats the purpose.

PR #645, opened 2023-08-08
Linux shell August 2023

eglerean:
Added page description and registration. Please review if you have time.


rkdarst:
Looks good, though I don't know of the times/etc other stuff.  We can work on updates for how to get terminal and prepare.

PR #651, opened 2023-09-16
scicomp/ssh: Updates throughout the SSH page

rkdarst:
- Generalize the page, since SSH config and keys are going to become
  much more important in the future.
- Better explain all the key options for connecting on different
  operating systems.

### DIFF
--- a/scicomp/ssh.rst
+++ b/scicomp/ssh.rst
@@ -1,190 +1,314 @@
 SSH
 ===
 
-This walk-through presumes that the user
+Secure Shell (SSH) is the standard program for connecting to remote
+servers and transferring data.  It is very secure and well-supported,
+so it's worth learning to use it properly.  This page both gives a bit
+of a crash course (top) and more details (bottom).
 
-- is working on a Linux machine, a mac or a windows machine with openssh installed (default on recent Windows 10+ versions)
-- has ``OpenSSH`` installed: ``ssh -V`` in the terminal to check
-- has an account on the server of interest
-- is connected to the Aalto network
+.. highlight:: console
 
-We will be focusing on connecting to *Triton* but the methods described below are applicable to any of Aalto's other :doc:`remote servers </aalto/remoteaccess>`.
+
+
+Setup
+-----
+
+Check the tabs below for your operating system and methods to see what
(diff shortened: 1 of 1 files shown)


rkdarst:
Ready for review, please review quickly

Gehock:
I've seen some recommendations about using ed25519 keys instead of RSA (for example, [github docs](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent)), although couldn't find any official guidance. If nothing else, ed25519 keys are nicer to handle because they're shorter (in terms of characters) than RSA. There's also a [list](https://www.aalto.fi/en/services/linux-shell-servers-at-aalto) or fingerprints of Aalto shell servers available.

scicomp/ssh.rst
```diff
@@ -260,55 +392,53 @@ and you would use ``ssh triton`` to log in. Any additional server configs can fo
...
 
 
 .. _example_config_for_ssh:
 
 Full sample config file
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-The following code is placed in the config file created above (i.e. ``~/.ssh/config`` on mac/linux or ``%USERPROFILE%.ssh/config`` on windows) ::
+The following code is placed in the config file created above
+(i.e. ``~/.ssh/config`` on Mac/Linux or ``%USERPROFILE%.ssh/config``
```
Gehock, line 415:
```suggestion
(i.e. ``~/.ssh/config`` on Mac/Linux or ``%USERPROFILE%/.ssh/config``
```
Fixing a typo

rkdarst:
> I've seen some recommendations about using ed25519 keys instead of RSA (for example, [github docs](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent)), although couldn't find any official guidance. If nothing else, ed25519 keys are nicer to handle because they're shorter (in terms of characters) than RSA.

I thought of that, but wondered, could it be there are still some common machines around that don't support ec25519.  At least Triton does so that's good.  With the decommissioning of taltta maybe everything does?

PR #677, opened 2023-10-20
aalto/jupyterhub-instructors/nbgrader-hints: Note about grade scaling

rkdarst:
- Note a feature that's added to not scale grades to 100 when
  exporting.
- Also note that the email conversion isn't needed usually.
- Review: automerge


Gehock:
LGTM, glad I happened to notice this PR just before I was about to start writing one with almost identical content.

PR #689, opened 2024-01-17
Generative ai tools

YuTian8328:
This PR adds a new page for available generative ai tools on triton.


rkdarst:
It's good to have a landing page like this.  Random thoughts:

* Introduction that somehow mentions the the ideas of "do it on the cloud" vs "do it locally batched from your code", and when you might use each of these.
* And maybe some about why you might want to use local models (cheaper for bulk processing, data security)?
* I'd recommend using `-` instead of `_` for the page name (matching other pages better)

then about main content:
* https://scicomp.aalto.fi/triton/apps/speech2text/ - this is the link to speech2text, which can be described separately from the main Whisper page (with emphasis on being usable by anyone)
* Is it worth mentioning the local LLM APIs we are working on?
* If you think it would fit, https://www.aalto.fi/en/services/aalto-gpt
* I wonder if a link to other info on using the general ChatGPT is worth it (or even is findable?)

I'll try to think of more later...

triton/apps/generative_ai_tools.rst
```diff
@@ -0,0 +1,17 @@
+Generative AI Tools on Triton
+=============================
+Speech to Text: Whisper
+----------------------------
+Whisper is a general-purpose speech recognition model. It is a multitasking model that can perform multilingual speech recognition, speech translation, and language identification. For the usage on Triton, please check out the `whisper <https://scicomp.aalto.fi/triton/apps/whisper/>`__ page.
```
rkdarst, line 5:
```suggestion
Whisper is a general-purpose speech recognition model. It is a multitasking model that can perform multilingual speech recognition, speech translation, and language identification. For the usage on Triton, please check out the :doc:`whisper <whisper>` page.
```
Internal links can be made this way, and it'll know when it breaks.  This is a relative path but also thinsg like /triton/apps/whisper or ../some-dir/page work.

rkdarst:
I sent some updates.  Still to do
- is the command line tool the thing I made?  It may not be worth mentioning (unless someone else thinks it has some benefit?)
- ensure I got the names of the LLM API correct
- rename page to use `-`

rkdarst:
Now the page has changed quite a bit and isn't just about Triton, but all AI tools.  It helps to guide people to the right kinds of resources.

Now that I think about it, the original page also could have its place - do we need that too, or is it OK for people to know the application they need and go straight to that page?

PR #692, opened 2024-02-14
Kickstart 2024 initial course page

rkdarst:


eglerean:
merging to release early, we revise later

PR #695, opened 2024-02-29
update instructions for openconnect

arashbm:
Cisco VPN now requires User Agent to contain the word AnyConnect to work.

https://scicomp.zulip.cs.aalto.fi/#narrow/stream/4-general/topic/VPN.20connection.20problem/near/276672

### DIFF
--- a/aalto/remoteaccess.rst
+++ b/aalto/remoteaccess.rst
@@ -70,8 +70,9 @@ Below is some quick reference info.
 -  Aalto windows: Start → Search → AnyConnect
 -  Personal Linux laptops: Use OpenConnect. Configuration on Ubuntu:
    Networks → Add Connection → Cisco AnyConnect compatible VPN. →
-   ``vpn.aalto.fi``. Then connect and use Aalto username/password. Or from
-   command line: ``openconnect https://vpn.aalto.fi``
+   Set Gateway to ``vpn.aalto.fi`` and User Agent to ``AnyConnect''.
+   Then connect and use Aalto username/password. Or from the command
+   line: ``openconnect https://vpn.aalto.fi --useragent=AnyConnect``
 -  Personal mac: `use Cisco AnyConnect VPN
    Client <https://download.aalto.fi/staff/>`__
 -  personal windows: `use Cisco AnyConnect VPN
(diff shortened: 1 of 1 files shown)


tpfau:
When I try to do this, I don't get to the SSO handler, which would be needed to log in. 
How do you set up the SSO handler for openconnect?

arashbm:
On Ubuntu 23.10 package `network-manager-openconnect-gnome` has a dependency on `libwebkit2gtk-4.0-37` and just opens up a webkit when connecting to the VPN. Seems like invoking from CLI has a [known issue](https://gitlab.com/openconnect/openconnect/-/issues/425) though. Should I just remove mentions to the CLI?

tpfau:
I think the problem is in exactly this package. Ubuntu 22.04 `network-manager-openconnect-gnome` is still at version 1.2.6-4, which (I think) does not have the fix setting AnyConnect as client ID for cisco vpn servers, and thus fails... 
I'm currently waiting for April to upgrade and use the next lts which hopefully has solved this.

arashbm:
At least on 1.2.10-1 the user agent still has to be set manually.

tpfau:
But how do you set it at all for the UI version? 
In my network manager, there is no optio to set the user Agent anywhere.

arashbm:
It seems to be there in the newer version:
image

tpfau:
Ok, well... I'll be waiting till 24.04 then since 22.04 doesn't seem to get that upgrade. :/

rkdarst:
Is there any reason to not merge this right now?

aalto/remoteaccess.rst
```diff
@@ -70,8 +70,9 @@ Below is some quick reference info.
 -  Aalto windows: Start → Search → AnyConnect
 -  Personal Linux laptops: Use OpenConnect. Configuration on Ubuntu:
    Networks → Add Connection → Cisco AnyConnect compatible VPN. →
-   ``vpn.aalto.fi``. Then connect and use Aalto username/password. Or from
-   command line: ``openconnect https://vpn.aalto.fi``
+   Set Gateway to ``vpn.aalto.fi`` and User Agent to ``AnyConnect''.
+   Then connect and use Aalto username/password. Or from the command
+   line: ``openconnect https://vpn.aalto.fi --useragent=AnyConnect``
```
rkdarst, line 75:
```suggestion
   line: ``openconnect https://vpn.aalto.fi --useragent=AnyConnect``.  This requires a new enough OpenConnect, at least Ubuntu 23.10 has it.
```

rkdarst:
I'm just merging this, it's better than before.  I'm sorry it took so long - thanks for the info!

PR #701, opened 2024-03-22
adding ttt4hpc

eglerean:


eglerean:
Merging. There is an issue with triton/apps/python-conda.rst that I will solve in another PR

PR #704, opened 2024-03-26
conf: Remove next and previous buttons

rkdarst:
- Not labeled descriptively, also most of the site isn't designed for
  scolling forward and backwards.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -197,6 +197,7 @@ else:
 html_theme_options = {
     'display_version': False,
     'navigation_depth': 5,
+    'prev_next_buttons_location': None,
     #'canonical_url': 'https://scicomp.aalto.fi/'
     }
 html_context.update(


tpfau:
Is there any possibility that we can have these on the Tutorials pages? 
Or should we do this during a large refactor, where we potentially move different parts of the docs to different locations, which are also compiled differently?

rkdarst:
I don't think they can be page-specific (but also, at least one tutorial has a subpage that shouldn't be included).  I have tried to add an explicit "what's next" section at the bottom of each page to move people towards the next tutorial (with a description about what is next), which I think should be OK-ish.

rkdarst:
I'm making an executive decision to merge these.

I just made sure all the links between tutorials are correct (the "what's next?" sections)

PR #706, opened 2024-04-05
Updating anaconda -> scicomp-python-env in triton documentation

tpfau:
This is a PR that will update references to anaconda to scicomp-python-env.

NOTE: Please do not merge before Triton v3 is in production
NOTE2: This PR does NOT update references to anaconda in the non triton specific instructions of the doc, since those also refer to workstations etc. We should probably work towards harmonization here, i.e. having the workstations (and thus aalto linux) be set up to use a new module, otherwise we start getting out of sync.


rkdarst:
Can you also update the cheatsheets (findable as a link in google drive)?

rkdarst:
Should we do this now?

PR #709, opened 2024-05-06
Clarifying use of python installation

tpfau:
This updates the python page to provide a separation on installing packages. 
No installation should happen with scicomp python env. If you need more packages -> use mamba
Separation is:
"General python" : module load scicomp-python-env
"Need more packages": Write an environment file and use module load mamba

### DIFF
--- a/triton/apps/python.rst
+++ b/triton/apps/python.rst
@@ -81,7 +81,7 @@ for you to port your code to another system.
    We stringly recommend not to instal packages using ``pip install --user``.
    If you do this, the package will be shared among all
    your projects, and will even overwrite any package installed in an environment.
-    It is quite likely that eventually, you will get some
+   It is quite likely that eventually, you will get some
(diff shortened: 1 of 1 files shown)


triton/apps/python.rst
```diff
@@ -66,23 +67,24 @@ Quickstart
...
 
 .. warning:: ``pip install --user`` can result in incompatibilities
 
-   If you do this, then the module will be shared among all
-   your projects. It is quite likely that eventually, you will get some
-   incompatibilities between the Python you are using and the modules
+   We stringly recommend not to instal packages using ``pip install --user``.
+   If you do this, the package will be shared among all
+   your projects, and will even overwrite any package installed in an environment.
+    It is quite likely that eventually, you will get some
```
rkdarst, line 84:
```suggestion
   It is quite likely that eventually, you will get some
```

rkdarst (approved):
good, let's do this ASAP

PR #712, opened 2024-05-21
triton: remove VDI as a connection recommendation

rkdarst:
- OOD now has graphics +
- It's still mentioned as a remote data access method which seems OK
- It's still mentioned in `triton/tut/interactive.rst` but that
  section needs revising later anyway
- review: automerge 3day if no other comments


tpfau:
This looks good to me. 
Suggestion from my side:
We can, in the re-vamp build a page dedicated to interactive/graphical use. Where we can mention that vdi is another possibility (if you don't want to use triton desktop.  But it's not the recommended way (Which would be Triton desktop), and we will not actively support it.

PR #715, opened 2024-05-29
Alternative example for array jobs with hardcoded parameters, chosen from array instead of case statement

AbsoluteVacuum:


eglerean:
Looks good to me. I would like the teachers of this part of next week's course to be aware of this change since it will be on the course pages so explicit ping to @rkdarst and @simo-tuomisto

rkdarst:
This is good, though part of the reason for the `case` example was that it would be completely obvious to someone looking at it (the goal is more to show how it works to someone that has never seen a shell script, than provide the most recommended solution).  I'd rather keep the page shorter and less advanced.

But we *should* have more advanced and cool array examples somewhere.  This would be good there (and maybe we should move another example or two from the page there, too?)  An `array` directory here?: https://github.com/AaltoSciComp/hpc-examples/

PR #728, opened 2024-07-26
Add an icon for external links

rkdarst:
- This uses the unicode up right arrow symbol.

- Unicode doesn't have an "external link icon" but up-right-arrow was
  suggested as a replacement.

- This works but the icon isn't distinguished as a separate thing from
  text, thus things such as speech to text would get messed up by it.
  Does this actually reduce accessibility more than it improves
  usability?

- A quick check didn't reveal how these should be tagged to not be
  read/the meaning be understood.


rkdarst:
Now supported by the theme natively.  It was enabled and now I'm closing this.

PR #737, opened 2024-08-28
More data updates

tpfau:
Added some links in the data index page. 
Could probably be even more in the table but before I add those I would like to know if you think this is a good idea.


rkdarst:
Looks good.  Made some minor updates to the text:

> You *really should* talk to your group leader about data storage
locations - data is important and your group leader needs to be aware
of where all the group's data is stored so they can coordinate all the
use.

I think it more than a second opinion: group leaders need to know what's going on, it should be one of the base things that is discussed any time a project starts.

PR #739, opened 2024-08-29
Fix Azure OpenAI API link

hsnfirooz:
The ITS page is confusing and hard to find.
The updated link has all the information (from how to request to API usage, costs, endpoints etc). The previous link doesn't provide the API information, just how to request.

(Although the structure of the ITS page is confusing, I think having this link is better).


rkdarst:
Thanks - I usually try to only link to open pages, but if this is the best that exists...

PR #740, opened 2024-08-29
URL changed for one lesson (sorry)

bast:
when renaming it I assumed the URL would forward but I was wrong


bast:
Context for rename: the name was too specific and we plan to reuse and evolve material for another course.

PR #749, opened 2024-10-28
Update python-for-scicomp-2024.rst

eglerean:


rkdarst:
Looks good (as far as I can tell).  Go?

eglerean:
I wanted to add the updated schedule, but it will have to wait

PR #752, opened 2024-11-07
removed a section that seemed spurious

susamerz:
I notice a section in the documentation of jupyterhub through OOD that seemed a) at least somewhat spurious, b) the example seemed to be using modules that are no longer there, and therefore confusing. Currently this is removing the whole section, I'd be happy to rework, if someone thinks this should stay

### DIFF
--- a/triton/apps/jupyter.rst
+++ b/triton/apps/jupyter.rst
@@ -246,29 +246,25 @@ The disadvantage is that you will need to create a kernel, the advantage is that
 is to use the existing R installations on Triton.
 
 .. tabs::
-
+ 
   .. tab:: Using a conda environment
      
         You will need to create your own conda environment with all packages that are necessary
-        to deploy the environment as a kernel.::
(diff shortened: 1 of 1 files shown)


rkdarst:
`r-irkernel/1.1-python3` is probabyl old, the thing we had before.  I do wonder, when did we ever need that section.  Maybe it's been replaced with the one below.

The section below also seems to have some problems and probably needs testing.  Maybe sometime we could look at it together - it would be good for someone else to know the R part and give it a good check, and right now I'm the main one who knows how jupyter and the kernels work...

Find me in garage and we can take a look...

rkdarst:
This was good!  I pushed something that adjusted various fromatting - it can be quite picky about all the indention and spacing.  I think it it's easier to look at my change than to try to explain everything individually.  (Text updates are minor and don't really matter that much.)

rkdarst (approved):

PR #758, opened 2025-01-29
Update Quota

hsnfirooz:
Fixed Home directory quota.

### DIFF
--- a/triton/usage/quotas.rst
+++ b/triton/usage/quotas.rst
@@ -3,7 +3,7 @@ Quotas
 ======
 
 Triton has quotas which limit both the space usage and number of files.
-The quota for your home directory (`$HOME`) is 20GB, for $WRKDIR by default is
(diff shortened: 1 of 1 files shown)


triton/usage/quotas.rst
```diff
@@ -3,7 +3,7 @@ Quotas
 ======
 
 Triton has quotas which limit both the space usage and number of files.
-The quota for your home directory is 10GB, for $WRKDIR by default is
+The quota for your home directory (`$HOME`) is 20GB, for $WRKDIR by default is
```
rkdarst, line 6:
```suggestion
The quota for your home directory (``$HOME``) is 20GB, for $WRKDIR by default is
```

our sphinx requires double backquotes (it can be configured to require only one, but single backquote can be other roles too such as internal links, and I choose the most portable choice when starting since it was easier to change to be more flexible later.)

PR #768, opened 2025-05-05
WIP: Update links for CSC

samumantha:
Updated the general CSC link to show our research pages (assumed more interesting for researchers) and link to our Course materials for info about connecting etc.

What exactly constitutes "active support"?

### DIFF
--- a/training/scip/kickstart-2025.rst
+++ b/training/scip/kickstart-2025.rst
@@ -77,8 +77,8 @@ students.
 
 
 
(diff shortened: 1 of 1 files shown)


rkdarst:
I guess "active support" meant someone who would be around and asking questions from the notes during the course.  But it can mean whatever you'd want it to mean (I probably wrote that to say "we can't promise someone can answer questions").

Should we merge this?

PR #777, opened 2025-06-02
kickstart-2025: New example cluster workflow

simo-tuomisto:
Added new intro section on cluster workflows, modified course timings


rkdarst (approved):
Looks good.  I do'nt have deep thoughts but it seems like just what we need to talk about.

PR #781, opened 2025-06-05
Remove the pytorch channel from the list of conda channels

rantahar:
The pytorch channel is deprecated and pytorch can be installed from conda-forge + nvidia instead.


rantahar:
- Add the new conda page and copy sections in. 
- Add an R example for the first language specific example.

rantahar:
- R examples added to all relevant sections
- Moved a note about pip in conda into a collapsible section. This is only about Python.
- Moved the previous `python-conda` section into a new `deprecated` folder

PR #783, opened 2025-06-11
about/science-it-history: New page with our background

rkdarst:
- To help maintain some institutional knowledge
- This first attempt is what I can quickly remember.  It needs good
  fact-checking and expansion by someone who was around for more of
  it.


eglerean:
This looks good to me, but since it is old I would like @rkdarst to confirm that it is good to merge.

rkdarst:
I recently chatted with someone about it and I think it's reasonably correct.  It can always be updated later - perhaps best to merge and let others correct it.

PR #784, opened 2025-09-02
Update docs to new GPU request syntax

simo-tuomisto:
This PR modifies the docs to reflect the new recommended way of requesting GPUs:

- `--gpus=1` for requesting a GPU.
- `--gpus=h200:1` for requesting specific GPU.
- `--gpus=1 --gres=gpu-vram:40g` for requesting GPUs with a certain minimum amount of VRAM.
- `--gpus=1 --gres=min-cuda-cc:80` for requesting GPUs with a certain minimum CUDA compute capability.

Examples have been changed to reflect the recommend way of requesting GPUs.


triton/ref/gpu.rst
```diff
@@ -2,19 +2,20 @@
...
-   NVIDIA H200(*)       | ``gpu-h200-141g-ellis``, ``gpu-h200-141g-short``  | ``hopper``              | ``h200``                                | 112            | gpu[50-63]   | Hopper         |                           | 141GB             | 9.0
-   AMD MI100 (testing)  |   *Not yet installed*              | ``mi100``                              | Use ``-p gpu-amd`` only, no ``--gres``  |                | gpuamd[1]    |                |                           |                   |
+   GPU brand name       | GPU name in Slurm (``--gpus=NAME:n``) | Amount of VRAM                   | CUDA compute capability         | total amount   | nodes            | GPUs per node | Compute threads per GPU   | Slurm partition (``--partition=``)               |
+   NVIDIA H200(*)       | ``h200``                               | 141GB (``--gres=gpu-vram:141g``) | 9.0 (``--gres=min-cuda-cc=90``) | 112            | gpu[50-63]       | 8             | 16896                     | ``gpu-h200-141g-ellis``, ``gpu-h200-141g-short`` |
```
rkdarst, line 6: I might leave the full command lines (``--gres=gpu-vram:141g``) out from this table and instead have another table below that goes into detail about the command line options.  We need special instructions that only the last one is used anyway.  The table can get quite long and right now the HTML doesn't scroll tables for some reason.

PR #816, opened 2026-04-07
draft AI agents on HPC

eglerean:


triton/usagepolicy.rst
```diff
@@ -77,6 +77,16 @@ these conditions of use:
...
 
 
+AI agents on Triton
+-------------------
+
+If you use AI coding agents (e.g. Claude Code, OpenAI Codex) on Triton or any other
+computer connected to our infrastructure, please read our policy on acceptable use,
+risks, and required cooperation:
+
+:doc:`AI Agents on HPC <AI_agents_on_HPC>`
```
rkdarst: Is there a more sustianable way to have this?  If we have a whole section here, there could as well be a section on not making too many small files, not making too short jobs, not opening /scratch in vscode, etc.  Having a way to highlight important things at account creation seems like a good idea, but how to avoid it having be a parallel doc index?

I can think of either a list of important things, or a generic point "please scan docs for pages that are relevant to your use".  We could also have a new box somewhere more prominent (triton docs page?  sidebar?) that says the currently relevant stuff that is new and non-obvious.  I had hoped that the Mastodon box could be this but we see how that went...

rkdarst:
todo: add to toctree.  I suggest under /triton/usage/

Suggested rename: /triton/usage/ai-agents.rst to match naming convention of other pages.  It will then be included in the toctree automatically.

triton/usagepolicy.rst
eglerean: I agree that maybe the link from usagepolicy is not the best since in the end this is not a policy, more about good practices. Should we move it under the list https://scicomp.aalto.fi/triton/#detailed-instructions ?

PR #822, opened 2026-05-12
draft of going public about local llms

eglerean:


aalto/llm-web-apis.rst
```diff
@@ -1,24 +1,96 @@
...
 ==================
 
-As a pilot service, :doc:`Aalto RSE </rse/index>` has a service
-running some common open-source LLMS (llama3, deepseek, etc.) available
-via the web.  This can be used for lightweight purposes via
-programming, but shouldn't replace batch usage (use
-:doc:`/triton/apps/llms`) or interactive chatting (use Aalto GPT).
+:doc:`Aalto RSE </rse/index>` runs an LLM gateway service at
+`https://llm-gateway.k8s.aalto.fi/ <https://llm-gateway.k8s.aalto.fi/>`__
+that is available to all Aalto users.  It serves several open-source
```
tpfau: Currently:
Staff/Employees. Students need to obtain a key from supervisors. 
'm open to allowing more general use.

aalto/llm-web-apis.rst
```diff
@@ -1,24 +1,96 @@
 Local LLM web APIs
 ==================
 
-As a pilot service, :doc:`Aalto RSE </rse/index>` has a service
-running some common open-source LLMS (llama3, deepseek, etc.) available
-via the web.  This can be used for lightweight purposes via
-programming, but shouldn't replace batch usage (use
-:doc:`/triton/apps/llms`) or interactive chatting (use Aalto GPT).
+:doc:`Aalto RSE </rse/index>` runs an LLM gateway service at
+`https://llm-gateway.k8s.aalto.fi/ <https://llm-gateway.k8s.aalto.fi/>`__
```
tpfau, line 5: The address (of the frontend) will change soon, but probably we can just keep it like this (there will be a forward from the backend to the frontend for non api related routes)
llm-gateway.aalto.fi will be the frontend site, but let's see.

aalto/llm-web-apis.rst
```diff
@@ -1,24 +1,96 @@
...
+- ``RedHatAI/gemma-4-31B-it-FP8-Dynamic``
+- ``google/gemma-4-31B-it``
+- ``google/gemma-4-E4B-it``
+- ``Qwen/Qwen3-VL-30B-A3B-Instruct-FP8``
+- ``Qwen/Qwen3-VL-30B-A3B-Thinking-FP8``
+- ``openai/gpt-oss-120b``
+
+All models support the OpenAI-compatible chat/completions API, so most
+existing OpenAI client libraries work without changes.
+
```
tpfau, line 85:
```suggestion
Only two models:
- ``Qwen/Qwen3-30B-A3B-Instruct-2507-FP8``
- ``RedHatAI/gemma-4-31B-it-FP8-Dynamic``
are permanently on, i.e. there is an instance of those running constantly. 
The other models are on demand, i.e. they will be spun up once a request comes in and 
will be turned off again if there are no requests for some time. 
Given the limited resources (at time of writing the whole supporting infrastructure has 8 L40s cards)
not all models can run at the same time and it is entirely possible that a model
will not spin up after a request because resources are in use.
```

aalto/llm-web-apis.rst
```diff
@@ -1,24 +1,137 @@
...
+     - Size (GB, estimated)
+   * - ``Qwen/Qwen3-30B-A3B-Instruct-2507-FP8``
+     - `Model card <https://huggingface.co/Qwen/Qwen3-30B-A3B-Instruct-2507-FP8>`__
+     - ~16
+   * - ``RedHatAI/gemma-4-31B-it-FP8-Dynamic``
+     - `Model card <https://huggingface.co/RedHatAI/gemma-4-31B-it-FP8-Dynamic>`__
+     - ~17
+   * - ``google/gemma-4-31B-it``
+     - `Model card <https://huggingface.co/google/gemma-4-31B-it>`__
+     - ~70
```
tpfau: Delete this for now, since the model doesn't want to start, multi-GPU model and there are some issues seemingly... 
```suggestion
```

aalto/llm-web-apis.rst
```diff
@@ -1,24 +1,137 @@
...
+     - ~62
+   * - ``Qwen/Qwen3-VL-30B-A3B-Thinking-FP8``
+     - `Model card <https://huggingface.co/Qwen/Qwen3-VL-30B-A3B-Thinking-FP8>`__
+     - ~16
+   * - ``openai/gpt-oss-120b``
+     - `Model card <https://huggingface.co/openai/gpt-oss-120b>`__
+     - ~70
+
+All models support the OpenAI-compatible chat/completions API, so most
+existing OpenAI client libraries work without changes. If you want to request new models, please email ``rse@aalto.fi``. Please note that only models smaller than ~70GB can be used in the current hardware we have.
```
tpfau: The system does NOT provide responses background functionality, i.e. submit now, come back later things. It does otherwise support responses.

aalto/llm-web-apis.rst
```diff
@@ -1,24 +1,138 @@
...
+   completion = client.chat.completions.create(
+       model="RedHatAI/gemma-4-31B-it-FP8-Dynamic",
+       messages=[
+           {"role": "system", "content": "Helpful assistant that writes python for research."},
+           {"role": "user", "content": "I need a python script to load a csv."}
+       ]
+   )
+
+   console = Console()
+   console.print(Markdown(completion.choices[0].message.content))
```
rkdarst, line 130: This is exactly the level of example I want to see!

aalto/llm-web-apis.rst
```diff
@@ -1,24 +1,138 @@
...
-:ref:`chat`.  That's also the best way to contact the developers
-(other contact methods are in :doc:`/help/index`.
 
-The API doesn't have it's own detailed documentation (ask us), but the
-API should be OpenAI compatible (for chat models) so many existing
-libraries work automatically. Here are some simple `examples. <https://github.com/AaltoSciComp/llm-examples/tree/main/aalto-llm-api>`__ 
+Who can use the endpoint and with what data?
+--------------------------------------------
+
+To use the local endpoint you need to have an active Aalto account and be part of Aalto staff. Students can get access via their course organiser (the organiser creates an API key for the students to use); teachers are responsible for revoking student keys once the course work is complete. The advantage of a fully local solution is that the endpoint can be use with data classified as confidential. Secret data should not be processed with these tools.
```
rkdarst: "Aalto staff" -> "Aalto employee" ?
To make it clear that it's not just service staff, or some other more strict interpretation of staff.

rkdarst (commented):
Looks good though I didn't read it in as much detail as the last one.  Two comments:

- Only available inside the Aalto network.
- Comment that it is free but also no resource sharing so no guaranteed level of service, this will be adjusted if demand increases.

eglerean:
Suggestions above have been added.

PR #824, opened 2026-06-03
PyTorch container example for ARM GPUs

Vinye:
I had this on my backlog. I'd like @simo-tuomisto to review this when you have the time.

Especially I'm thinking about the pyyaml part, pip will warn about running as root. I just solved it by pinning pyyaml instead of creating a venv inside the container. When I looked at [the TTT4HPC lesson example](https://coderefinery.github.io/hpc-containers/gpus/#example-container-model-training-with-accelerate) where you install `transformers`, it doesn't mention this pyyaml thing. @simo-tuomisto, what do you think would be the best approach here?

Maybe @rkdarst could also take a look at the style in general? It is the first time I write a little longer tutorial here.

### DIFF
--- a/triton/usage/gracehopper.rst
+++ b/triton/usage/gracehopper.rst
@@ -76,7 +76,7 @@ The hello-world.cu could be e.g.::
     return 0;
   }
 
-Building a simple pytorch environment
+Building a simple PyTorch environment
 -------------------------------------
 
 Nvidia provides ARM containers for PyTorch, which you can use as a starting point for your own containers.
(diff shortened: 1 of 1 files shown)


triton/usage/gracehopper.rst
```diff
@@ -76,3 +76,95 @@ The hello-world.cu could be e.g.::
     return 0;
   }
 
+Building a simple pytorch environment
```
Vinye: Nit: PyTorch

simo-tuomisto:
Looks great!

Vinye:
Taking the liberty to merge this

PR #835, opened 2026-09-08
Fix typo in 'Storage' section header

nikosavola:


eglerean:
Thank you for the fix! If you are around CS IT office and need a cookie, you earned at least a handful of them. :)

PR #836, opened 2026-09-15
gpu interactive limits

hsnfirooz:
I've added our policy on using interactive jobs with GPUs.


triton/tut/interactive.rst
```diff
@@ -178,6 +178,12 @@ then sshes to there and connects to the shell::
   slowly when sent across the general Internet.
 
 
+Interactive shell with GPU nodes
+-------------------------------
```
rkdarst:
```suggestion
--------------------------------
```

triton/tut/interactive.rst
```diff
@@ -178,6 +178,12 @@ then sshes to there and connects to the shell::
...
+
+For interactive jobs with gpu resources, it is advised to use `gpu-debug` partition. `gpu-debug` has a higher priority and your job will be scheduled faster.
```
rkdarst:
```suggestion
For interactive jobs with gpu resources, it is advised to use ``gpu-debug`` partition. ``gpu-debug`` has a higher priority and your job will be scheduled faster.
```

triton/tut/interactive.rst
```diff
@@ -178,6 +178,12 @@ then sshes to there and connects to the shell::
   slowly when sent across the general Internet.
 
 
+Interactive shell with GPU nodes
+--------------------------------
+
+For interactive jobs with gpu resources, it is advised to use ``gpu-debug`` partition. ``gpu-debug`` has a higher priority and your job will be scheduled faster.
```
rkdarst:
```suggestion
For interactive jobs with GPU resources, it is advised to use ``gpu-debug`` partition. ``gpu-debug`` has a higher priority and your job will be scheduled faster.  Batch GPUs are designed mainly for efficient batch computation and jobs which appear to be interactive and have low GPU usage may be cancelled with or without notice.
```

triton/tut/interactive.rst
```diff
@@ -178,6 +178,12 @@ then sshes to there and connects to the shell::
...
+
+If you need a specific GPU type, you can submit your interactive job to other partitions too, but keep the time limit of the job to 4 hours. Also the number of GPUs requested in an interactive job should be kept to 1 GPUs.
```
rkdarst:
```suggestion
If you need a specific GPU type, you can submit your interactive job to other partitions too, but keep the time limit of the job to 4 hours. Also the number of GPUs requested in an interactive job should be kept to 1 GPUs. Limited use of batch GPUs may be accepted if the reservation is kept to when you are actively at the computer using it, but frequently save because we can't guarantee your job won't be killed.
```

rkdarst (commented):
Some wording suggestions.

PR #837, opened 2026-09-16
updating model table + legal disclaimer

eglerean:


eglerean:
ah I should change the examples to use qwen

